PR #50, opened 2025-02-21
🚧WIP ✨ Add support for json defined agents

rparcus:


rparcus:
It works! And it dynamically adds a module that we can call in elixir.
image

z-spectral (approved):

PR #51, opened 2025-02-24
Pure python signal handler

chitacan:
Updated to allow configuring pure python prism in agent's signal handlers.

We can define pure python prism based on `Prism` abstract python class:

```py
from lux.prism import Prism

class HelloPrism(Prism):
    id='aaace2b1-8089-4d4e-b68e-2ce904da12f0'
    name='HelloPrism'
    description='A simple hello world prism'

    def handler(self, signal, context):
        # your own prism logic
        return {
            'message': f'Hello, {signal.get("name", "world")}!'
        }
    def new():
        return HelloPrism()
```

use defined python prism in the agent like this:

```ex
defmodule TestAgent do
  use Lux.Agent,
    name: "test agent",
    signal_handlers: [
      {SignalSchema, {:python, "/path/to/python_prism.py"}}
    ]
end
```

We can also define python prism in the agent module directly. (Not recommended now because of elixir formatting. but it's possible :))

```ex
defmodule TestAgent2 do
  use Lux.Agent,
    name: "test agent2",
    signal_handlers: [
      {
        TestSchema,
        {:python,
         """
         from lux.prism import Prism

         class HelloPrism(Prism):  
             id='aaace2b1-8089-4d4e-b68e-2ce904da12f0'
             name='HelloPrism'
             description='A simple hello world prism'

             def handler(self, signal, context):
                 return {
                     'message': f'Hello, {signal.get("name", "world")}!'
                 }
             def new():
                 return HelloPrism()
         """}
        }
    ]
end
```

Caveats and notes:

  * In theory, users can define python prism anywhere in the file system. However, If user define files outside of project, it may not be included in the output when user create a release.
  * The environment in which prism runs only supports the `venv` set in lux.


priv/python/lux/prism.py
```diff
@@ -0,0 +1,28 @@
...
+
+class Prism(ABC):
+    def __init__(self, **kwargs):
+        self.id = kwargs.get('id', str(uuid.uuid4()))
+        self.name = kwargs.get('name', '')
+        self.description = kwargs.get('description', '')
+        self.input_schema = kwargs.get('input_schema', {})
+        self.output_schema = kwargs.get('output_schema', {})
+    
+    def get_metadata(self):
```
rparcus, line 13: Prisms have the `view/0` function in elixir, which will return the Prism struct. Could we stick to that? We could return a python Dict., that will be translated to a Prism Struct automatic if returned back to elixir.

test/support/python_prism.py
```diff
@@ -0,0 +1,15 @@
+from lux.prism import Prism
+
+class SimplePrism(Prism):
```
rparcus, line 3: Could we have a test that verifies that we are able to get the whole Prism struct out of a python defined prism? This will be very important for example when feeding this prism to an LLM as tool for tool calling: description, schema, etc all need to be there.

test/support/python_prism.py
```diff
@@ -0,0 +1,15 @@
...
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
+    def handle_signal(self, signal, context):
+        return {
+            'message': f'Hello, {signal.get("name", "prism")}!',
+        }
+
+
+def create_prism():
```
rparcus, line 13: Is this a function that would need to be implemented? Could we write this module as an example of a python prism in the way that others should write them? In other words, it should not have test code or "factory" stuff in it, and it should resemble a real prism that someone would write in python.

Also, we need to make sure that loading prisms with dependencies works out of the box.

test/support/python_prism.py
rparcus: In this case, we import Prism, so that should already work...

test/support/python_prism.py
chitacan: > Is this a function that would need to be implemented?

Unfortunately, yes.. For now we need a function to handle the Prism class instantiation.
Without this, I couldn't find a good place to define static values (e.g. id)

> Also, we need to make sure that loading prisms with dependencies works out of the box.

This will work like python module in `priv/python`, I'll add more test cases.

test/support/python_prism.py
chitacan: I will cover this part in the next PR.

chitacan:
@rparcus Found some issues our custom `Atom` implementation and erlport's `Atom`.

* our custom `Atom` class cannot encode & decode atom type properly. Looks like it works on python test, but on the elixir side we got wierd `{:"$erlport.opaque", :python, <<...>>}` tuple.
CleanShot 2025-02-25 at 17 06 29@2x
CleanShot 2025-02-25 at 17 06 48@2x

* `erlport` 0.6.0 version is not compatible with python 3. Whenever I tried to import the module, I got following error message.
CleanShot 2025-02-25 at 17 08 50@2x

I upgraded [erlport version to 0.9.8](http://erlport.org/downloads/#id4) and now the `Atom` class works as expected. However we have to use `Atom` class from `erlport.erlterms` module.

CleanShot 2025-02-25 at 17 15 48@2x
CleanShot 2025-02-25 at 17 16 32@2x

test/unit/lux/prism_test.exs
```diff
@@ -117,4 +117,20 @@ defmodule Lux.PrismTest do
...
+        |> Enum.concat(["..", "..", "support", "simple_prism.py"])
+        |> Path.join()
+        |> Path.expand()
+
+      prism = Lux.Prism.view(simple_prism)
+      assert %Prism{} = prism
+      assert prism.name == "Simple Prism"
+      assert prism.description == "A very simple prism"
+    end
+  end
```
rparcus, line 135: What happens if we try to define a prism entirely inside a python eval block, so there is no reference to a file path? ✨

chitacan: Yes, in that case, user should pass entire python code block;;;

test/support/python_prism.py
rparcus: This is amazing!

test/support/simple_prism.py
```diff
@@ -0,0 +1,34 @@
+from lux.prism import Prism
+
+class SimplePrism(Prism):
```
rparcus, line 3: Perfect! Great job!

priv/python/erlport/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+"""Erlang port protocol."""
+
+__author__ = "Dmitry Vasiliev <dima@hlabs.org>"
+__version__ = "1.0"
+
+from erlport.erlterms import Atom, List, ImproperList
```
rparcus, line 33: 👍

rparcus (approved):
I think that this is ready to go. Will you work on docs on another PR?

PR #122, opened 2025-02-28
Anthropic llm client

rparcus:
I was creating an issue for this but Claude one-shotted the implementation while I wasn't looking. 😅


lux/test/unit/lux/llm/anthropic_test.exs
```diff
@@ -0,0 +1,159 @@
...
+      # Test implementation for token caching
+      # This would verify that the token cache is working correctly
+      # and improving performance
+    end
+
+    test "supports load balancing" do
+      # Test implementation for load balancing
+      # This would verify that requests are distributed across
+      # multiple API keys
+    end
```
rparcus, line 157:
```suggestion
```

rparcus:
I don't have a dedicated anthorpic key so didn't add integration tests actually hitting their servers. We should have those before this goes into main.

PR #128, opened 2025-03-02
feat: Lux storage module including postgres adapter

IdrisHanafi:
# What's new?

Introducing the Lux storage module that enables postgres DB connections within your lux agent.

## Usage and Testing:
```
docker run -it -p 5432:5432 -e POSTGRES_PASSWORD=postgres -e POSTGRES_DB=lux_dev postgres

cd lux # within the repo
mix ecto.migrate -r Lux.Repo
mix test
```


lux/lib/lux/storage.ex
```diff
@@ -0,0 +1,152 @@
+defmodule Lux.Storage do
```
rparcus, line 1: This is basically just a Ecto.Repo wrapper, so the question comes to mind regarding wether we could call `Lux.Repo` `Lux.Storage` instead. (friendlier to non elixir devs?). 

Provinding `Lux.Repo` is enough and we don't need this other level here.
Any LLM would know very well what to do with Lux.Repo.

lux/test/lux/storage/job_opening_storage_test.exs
```diff
@@ -0,0 +1,176 @@
+defmodule Lux.Storage.JobOpeningStorageTest do
+  use ExUnit.Case, async: true
```
rparcus, line 2: In the main `Lux` project we have `UnitCase` and `IntegrationCase`, so that only the right tests run when we call something like `mix test.unit`. This is probably not necessary for `LuxApp`.

lux/test/lux/storage/schemas/job_opening_test.exs
```diff
@@ -0,0 +1,103 @@
+defmodule Lux.Storage.Schemas.JobOpeningTest do
+  use ExUnit.Case, async: true
+  alias Lux.Storage
+  alias Lux.Storage.Schemas.JobOpening
+  alias Lux.Repo
+
+  setup do
+    # Start the sandbox for this test process
+    Ecto.Adapters.SQL.Sandbox.mode(Lux.Repo, :manual)
```
rparcus, line 9: Usually we inlcude this sandbox code in a module called DataCase, and use that as the test template that loads all helpers and what not.

IdrisHanafi:
> @IdrisHanafi !!! Thanks for the PR! Could you move the database stuff into LuxApp (under `./lux_app` project)? We already have a database there, as well as a phoenix server. I think that the part about managing jobs should be handled there instead of in the core lux project.

Agreed! I had a feeling that's where it should've lived. I originally implemented this thinking that we would have a variety of storage adapters out of the box, but that's a discussion for next time. The next iteration would be purely to just serve as a "persistent" memory.

In the meantime, I'll close this PR and re-open once i revisit--we have an agentic hedge fund to cook first. 🫡

Issue #141, opened 2025-03-04
Outdated Lux versions in Livebook guides


Ian-GL:
Many of the livebook guides are currently referencing older Lux versions like `0.2.0` and `0.4.0` instead of the latest version.

PR #142, opened 2025-03-04
Update Lux version in livebook guides

Ian-GL:
This updates the livebook guides to use the latest release `0.5.0`. Fixes https://github.com/Spectral-Finance/lux/issues/141.


lux/guides/agents.livemd
```diff
@@ -3,7 +3,7 @@
 ```elixir
 Mix.install(
   [
-    {:lux, "~> 0.4.0"},
+    {:lux, "~> 0.5.0"},
```
chitacan, line 6: How about use `>=` operator? There is no need to modify this even if a new minor version comes out later.

```suggestion
    {:lux, ">= 0.5.0"},
```

chitacan (approved):
Thanks for do this!!

lux/guides/agents.livemd
Ian-GL: Good point! I've changed it in https://github.com/Spectral-Finance/lux/pull/142/commits/57016e7bb5e9f0eec1430a20aa35735f6274d135

chitacan (approved):
Thanks @Ian-GL !!

PR #144, opened 2025-03-06
Update social links

Ian-GL:
Updated the links to socials and removed the "coming soon" of recently added guides.


rparcus (approved):
Thanks!

PR #148, opened 2025-03-07
Add livebook button

chitacan:
Added `Run in Livebook` buttons in every `*.livemd` files. They only can be rendered in the GitHub. The added button is only exposed on GitHub, not on Hex doc or livebook. Now lux users can easily access Livebook from both GitHub and Hex Doc.

Unfortunately, I couldn't completely unify the position of the button. In Livebook, due to the Markdown rendering constraints, there was a problem where the structure of the document changed when another element was inserted between h1 and the first code block, so I had to place the button under h2.

| GitHub | Hex Doc | Livebook |
|--------|--------|--------|
| CleanShot 2025-03-07 at 15 32 53@2x | CleanShot 2025-03-07 at 15 33 39@2x | CleanShot 2025-03-07 at 15 36 42@2x |

### DIFF
--- a/lux/guides/agents.livemd
+++ b/lux/guides/agents.livemd
@@ -34,11 +34,13 @@ Application.put_env(:venomous, :snake_manager, %{
 Application.ensure_all_started([:lux, :ex_unit, :kino])
 ```
 
-## Section
+## Overview
+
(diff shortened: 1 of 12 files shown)


rparcus:
We're on the way to have the best OSS docs ever. Thanks @chitacan!

rparcus (approved):

PR #126, opened 2025-02-28
Etherscan data integration

DicksonWu654:

PR #158, opened 2025-03-10
feat: Add protocol TVL prism implementation (part of #88)

chainparser:
@rparcus For your perusal, please.
Trying to know whether I'm on the right path here.  

```
# Start interactive Elixir
iex -S mix

# Test with valid protocol
iex> Lux.Prisms.TvlPrism.run(%{protocol: "uniswap"})

# Test error case
iex> Lux.Prisms.TvlPrism.run(%{protocol: "invalid-protocol"})
```

### DIFF
--- a/lux/lib/lux/prisms/protocol_tvl_prism.ex
+++ b/lux/lib/lux/prisms/protocol_tvl_prism.ex
@@ -1,0 +1,96 @@
+defmodule Lux.Prisms.TvlPrism do
+  @moduledoc """
+  A prism that checks Total Value Locked (TVL) of a DeFi protocol using DeFiLlama API.
+
+  ## Examples
+
+      iex> Lux.Prisms.TvlPrism.run(%{protocol: "uniswap"})
+      {:ok, %{
+        tvl: 3654321098.76,
+        protocol: "uniswap"
+      }}
+  """
+
+  use Lux.Prism,
+    name: "TVL Checker",
+    description: "Checks Total Value Locked (TVL) of a DeFi protocol",
+    input_schema: %{
+      type: :object,
+      properties: %{
+        protocol: %{
+          type: :string,
+          description: "Protocol slug from DeFiLlama"
+        }
+      },
+      required: ["protocol"]
+    },
+    output_schema: %{
+      type: :object,
+      properties: %{
+        tvl: %{
+          type: :number,
+          description: "Total Value Locked in USD"
+        },
+        protocol: %{
+          type: :string,
+          description: "Protocol slug used for query"
+        }
+      },
+      required: ["tvl", "protocol"]
+    }
+
+  import Lux.Python
+
+  require Lux.Python
+
+  def handler(%{protocol: protocol} = input, _ctx) do
+    with {:ok, %{"success" => true}} <- Lux.Python.import_package("requests"),
+         {:ok, result} <- get_tvl(protocol) do
+      {:ok, atomize_keys(result)}
+    else
+      {:ok, %{"success" => false, "error" => error}} ->
+        {:error, "Failed to import requests: #{error}"}
+
+      {:error, reason} ->
+        {:error, reason}
+    end
+  end
+
+  defp get_tvl(protocol) do
+    result =
+      python variables: %{protocol: protocol} do
+        ~PY"""
+        import requests
+
+        def fetch_tvl(protocol):
+            url = f"https://api.llama.fi/tvl/{protocol}"
+            try:
+                response = requests.get(url, timeout=10)
+                response.raise_for_status()
+                return {
+                    "tvl": float(response.text),
+                    "protocol": protocol
+                }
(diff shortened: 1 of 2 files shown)


rparcus (changes requested):
Hi @chainparser,

Thanks for experimenting before diving deeper. Great approach! Also, nice job discovering the Python interface! (Looks like you've been reading the docs! 📖👀)

For a functionality like this, where we're simply loading data from an external resource, it's better to use the Lux.Lens abstraction instead of Prism.

There are guides and tests (unit and integration) that cover Lux.Lens, and you'll see that it handles everything for you. Your main task is just to describe what the Lens is about so that an LLM can use it effectively.

If you need more references, you can find [dozens of examples](https://github.com/Spectral-Finance/lux/pull/126/files#diff-55aabf4a8ad8ee0acb7c16f7027fe9447d4b56600a43f5ce5981594449f16a84) in this open PR.

Another thing to keep in mind: If you're introducing capabilities that depend on third-party services, it's important to provide an easy way for developers to set up API keys and configurations.

For that, please check out:

- `Lux.Config`
- `config/runtime.ex`
- The `.*.envrc` files

I hope this helps. Thanks again for checking in! Looking forward to seeing the updates. 🚀

chainparser:
@rparcus, it seems like creating a Lens of this nature is a bit tricky. Why do I say that? The URL specified in the Lens is static and can only be modified by appending the protocol to the URL in this manner: `https://api.llama.fi/tvl?protocol=aave`, while we want the protocol appended as `https://api.llama.fi/tvl/aave`. As far as I know, I have not found a way of building the URL like the second example. 

Here's how my code is looking now:

```elixir
# lib/lux/lenses/defi_llama_lens.ex

defmodule Lux.Lenses.DefiLlamaLens do
  @moduledoc """
  Fetches Total Value Locked (TVL) data from DeFiLlama API.

  ## Parameters
  - protocol: Protocol slug from DeFiLlama (e.g., 'uniswap', 'aave')
  """

  use Lux.Lens,
    name: "DeFiLlama TVL",
    description: "Fetches Total Value Locked (TVL) data for DeFi protocols",
    url: "https://api.llama.fi/tvl/{}",
    method: :get,
    rate_limiting: [max_requests: 10, interval: 1000],
    schema: %{
      type: :object,
      properties: %{
        protocol: %{
          type: :string,
          description: "Protocol slug identifier"
        }
      },
      required: ["protocol"]
    }

  # @impl true
  # def before_focus(params) do
  #   case validate_params(params) do
  #     :ok -> {:ok, params}
  #     {:error, reason} -> {:error, "Invalid parameters: #{reason}"}
  #   end
  # end

  @impl true
  def after_focus(tvl) when is_number(tvl) do
    {:ok, %{
      tvl: tvl,
      raw_data: tvl,
      timestamp: DateTime.utc_now()
    }}
  end

  def after_focus(_response) do
    {:error, "Invalid TVL response format"}
  end

  # defp validate_params(%{"protocol" => protocol}) when byte_size(protocol) > 0, do: :ok
  # defp validate_params(_), do: {:error, "Protocol must be a non-empty string"}
end
```

rparcus:
> @rparcus, it seems like creating a Lens of this nature is a bit tricky. Why do I say that? The URL specified in the Lens is static and can only be modified by appending the protocol to the URL in this manner: `https://api.llama.fi/tvl?protocol=aave`, while we want the protocol appended as `https://api.llama.fi/tvl/aave`. As far as I know, I have not found a way of building the URL like the second example.
> 
> Here's how my code is looking now:
> 

Hi @chainparser. Thanks for raising this. It's a totally valid use-case and we should be able to cover for it with a small change to how `before_focus` works. 
I think that it should take the entire lens and return it, instead of exposing only the params. With that, we will be able to modify the url before doing the call.

chainparser:
@rparcus that's right. Have an Idea of the specific codeblock? Prefer to keep it pushing ASAP. My quick search tells me 
`/lux/lib/lux/lens.ex` at line `109`.

```Elixir 
def after_focus(body), do: {:ok, body}
def before_focus(params), do: params
```
PS: I also sent you a direct message on Discord. Same username as GitHub.

rparcus:
> @rparcus that's right. Have an Idea of the specific codeblock? Prefer to keep it pushing ASAP. My quick search tells me `/lux/lib/lux/lens.ex` at line `109`.
> 
> ```elixir
> def after_focus(body), do: {:ok, body}
> def before_focus(params), do: params
> ```
> 
> PS: I also sent you a direct message on Discord. Same username as GitHub.

Hey, that kinda it, but nor really. I started working on it but sadly have to refactor hundreds of tests... Will get the fix merged soon so you get unblocked.

chainparser:
Okay, sir. Very much appreciated.

chainparser:
@rparcus in the meantime, are there any docs I can read to understand how Lux works?

PR #159, opened 2025-03-10
Add topic attribute to Signal struct

Ian-GL:
Adding the optional `topic` attribute to the signal struct so the Lux servers implementations can use this field to broadcast the signal to any interested agents vs. having a single receiver.

### DIFF
--- a/lux/lib/lux/signal.ex
+++ b/lux/lib/lux/signal.ex
@@ -10,6 +10,7 @@ defmodule Lux.Signal do
           sender: String.t() | nil,
           recipient: String.t() | nil,
           timestamp: DateTime.t(),
+          topic: String.t() | nil,
           metadata: map(),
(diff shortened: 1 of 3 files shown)


lux/lib/lux/signal.ex
```diff
@@ -10,6 +10,7 @@ defmodule Lux.Signal do
           sender: String.t() | nil,
           recipient: String.t() | nil,
           timestamp: DateTime.t(),
+          topic: String.t() | nil,
```
rparcus, line 13: Even if early, could we also add a line about this in `lux/guides/signals.livemd`? We can mention that it's for an upcoming feature.

rparcus (dismissed):
💙

rparcus (approved):
Thanks for the docs too!

PR #162, opened 2025-03-13
✨implement simple message prism and lens

sojinmm:
Add Discord message signal and schema implementation

This PR introduces the initial Discord integration functionality with the following components:

- Add OutgoingMessageSchema for Discord message validation
  - Implements message content validation (max 2000 chars)
  - Validates required channel_id field
  - Following Discord API specifications for message creation

- Add DiscordMessage signal implementation
  - Basic signal structure for sending messages to Discord channels
  - Integration with OutgoingMessageSchema for payload validation

- Add comprehensive unit tests
  - Validate successful message creation with required fields
  - Test content length restrictions
  - Verify mandatory field validation

This is the first step towards broader Discord integration features. Future PRs will expand on this foundation with additional Discord-related functionality.


sojinmm:
I’ve written the schema as shown below, following the existing code. Since we’re using Ecto, would it be better to continue using this approach moving forward?

rparcus:
> I’ve written the schema as shown below, following the existing code. Since we’re using Ecto, would it be better to continue using this approach moving forward?

We are not using ecto, what do you mean?

sojinmm:
> > I’ve written the schema as shown below, following the existing code. Since we’re using Ecto, would it be better to continue using this approach moving forward?
> 
> We are not using ecto, what do you mean?

I misspoke earlier. I was actually trying to check if we should continue not using Ecto, since we’re not using it right now. Instead of focusing on schema definition, I’ll first implement simple prisms and lenses, make the adjustments, and then push the changes.

sojinmm:
```zsh
    error: def focus/2 defines defaults multiple times. Elixir allows defaults to be declared once per definition. Instead of:

        def foo(:first_clause, b \\ :default) do ... end
        def foo(:second_clause, b \\ :default) do ... end

    one should write:

        def foo(a, b \\ :default)
        def foo(:first_clause, b) do ... end
        def foo(:second_clause, b) do ... end

    │
 51 │   def focus(input, ctx \\ %{})  # Single default declaration
    │       ^
    │
    └─ lib/lux/lenses/discord/message_lens.ex:51:7
```
    
 I'm getting the following error during compilation, but I believe I've implemented it according to the required format. Could I have missed something?

lux/test/integration/discord_message_lens_test.exs
```diff
@@ -0,0 +1,270 @@
+defmodule Lux.Integration.DiscordMessageLensTest do
+  use IntegrationCase, async: true
+
+  alias Lux.Lenses.Discord.MessageLens
+
+  @test_channel_id "123456789012345678"
+
+  setup do
+    Req.Test.verify_on_exit!()
```
rparcus, line 9: If this is an integration test, we should not be using Req. Integration tests actually hit the main service we want to test the integration for.

lux/lib/lux/prisms/discord/message_prism.ex
```diff
@@ -0,0 +1,281 @@
...
+      {:error, %{type: "rate_limit", retry_after: 5000}}
+
+  The prism reads Discord authentication details from configuration:
+  - :discord_token - Discord bot token for authentication
+  """
+
+  use Lux.Prism,
+    name: "Discord Message Sender",
+    description: "Sends messages to Discord channels",
+    url: "https://discord.com/api/v10/channels/:channel_id/messages",
```
rparcus, line 64: It looks like you are setting a lot of parameters that would be present in Lenses here. `url`, `method`, etc are not attributes of Prisms.

rparcus: If you try to compile this code, you indeed get a warning about this 
image

lux/lib/lux/prisms/discord/message_prism.ex
```diff
@@ -0,0 +1,281 @@
...
+      iex> handler(%{"id" => "123", "content" => "Hello"}, _ctx)
+      {:ok, %{status: "success", message: %{id: "123", content: "Hello", ...}}}
+
+      iex> handler(%{"code" => 50001, "message" => "Missing Access"}, _ctx)
+      {:error, %{type: "discord_api_error", code: 50001, message: "Missing Access"}}
+  """
+  @impl true
+  def handler({:error, error}, _ctx), do: {:error, error}
+
+  def handler(response, _ctx) when is_map(response) do
```
rparcus, line 229: Prisms are not like lenses... you need to implement the logic yourself. For example, here you have to actually use Req to call discord.

lux/mix.lock
```diff
@@ -36,6 +41,7 @@
...
   "nodejs": {:hex, :nodejs, "3.1.2", "a4dc114102782dcdc5fc13989398d19e594dd4e3a69181c8a8d54d08f4f950a4", [:mix], [{:jason, "~> 1.0", [hex: :jason, repo: "hexpm", optional: false]}, {:poolboy, "~> 1.5.1", [hex: :poolboy, repo: "hexpm", optional: false]}, {:ssl_verify_fun, "~> 1.1.7", [hex: :ssl_verify_fun, repo: "hexpm", optional: false]}], "hexpm", "7b01eee72c16d919b122f7096918abe9d6062d3cc299afe694e72412bbe90c92"},
+  "nostrum": {:hex, :nostrum, "0.10.4", "a316d08b19104f34c5fd5aa56674907350899a5f0c2483afdf5586296bd0ce07", [:mix], [{:castle, "~> 0.3.0", [hex: :castle, repo: "hexpm", optional: false]}, {:certifi, "~> 2.13", [hex: :certifi, repo: "hexpm", optional: false]}, {:ezstd, "~> 1.1", [hex: :ezstd, repo: "hexpm", optional: true]}, {:gun, "~> 2.0", [hex: :gun, repo: "hexpm", optional: false]}, {:jason, "~> 1.4", [hex: :jason, repo: "hexpm", optional: false]}, {:mime, "~> 1.6 or ~> 2.0", [hex: :mime, repo: "hexpm", optional: false]}], "hexpm", "fcc2642bf5b09792865ec2c26c1a11c6aa5432bc623a65dd81141e1eab9f1b99"},
```
rparcus, line 44: Where are we using this? You could for example use it inside Prisms if it's better than writing your own simple http client code.

lux/lib/lux/lenses/discord/message_lens.ex
```diff
@@ -0,0 +1,193 @@
...
+      required: ["channel_id"]
+    }
+
+  require Logger
+
+  @doc """
+  Adds Discord bot token to request headers.
+  """
+  def add_auth_header(lens) do
+    token = Application.fetch_env!(:lux, :discord_token)
```
rparcus, line 113: Where are we setting this Application variable?

rparcus (changes requested):

PR #163, opened 2025-03-17
✨Channel Prism and Lens implementations

sojinmm:
This PR enhances the Discord Channel Prism and Lens implementations to better align with Lux guidelines and improve overall code quality.

Channel Prism:
- Implements CRUD operations for Discord channels (create, update, delete)
- Handles channel permission management
- Supports various channel types (text, voice, forum, etc.)
- Implements Discord API rate limiting handling
- Provides comprehensive input validation through JSON schema
- Handles Discord-specific error cases

Channel Lens:
- Provides read operations for Discord channels
- Supports both single channel and guild-wide channel listing
- Implements proper error handling for Discord API responses

Both implementations follow Discord API v10 specifications and handle:
- Discord Snowflake ID validation
- Bot token authentication
- Rate limit responses
- Channel-specific field validations
- Proper error formatting

### DIFF
--- a/lux/lib/lux/lenses/discord/channel_lens.ex
+++ b/lux/lib/lux/lenses/discord/channel_lens.ex
@@ -1,0 +1,138 @@
+defmodule Lux.Lenses.Discord.ChannelLens do
+  @moduledoc """
+  A lens for reading Discord channel information.
+
+  This lens provides functionality to:
+  1. Get channel information using GET /channels/{channel.id}
+  2. List channels in a guild using GET /guilds/{guild.id}/channels
+  3. Get channel permissions (included in both responses)
+
+  ## Authentication
+  Requires a Discord bot token in the application configuration:
+  ```elixir
+  config :lux, :discord_token, "your-bot-token"
+  ```
+
+  ## Rate Limiting
+  Discord API has rate limits per route. This lens handles rate limiting by:
+  - Respecting the retry_after header
+  - Following the global rate limit
+  - Using proper request headers
+
+  ## Examples
+      # Get a specific channel
+      iex> Lux.Lenses.Discord.ChannelLens.focus(%{
+      ...>   channel_id: "987654321"
+      ...> }, %{})
+      {:ok, %{channel: %{"id" => "987654321", "name" => "example-channel", ...}}}
+
+      # List channels in a guild
+      iex> Lux.Lenses.Discord.ChannelLens.focus(%{
+      ...>   guild_id: "123456789"
+      ...> }, %{})
+      {:ok, %{channels: [%{"id" => "987654321", "name" => "general", ...}]}}
+
+  ## Error Cases
+      # Invalid channel ID
+      iex> Lux.Lenses.Discord.ChannelLens.focus(%{channel_id: "invalid"}, %{})
+      {:error, %{type: "discord_api_error", code: 50004, message: "Invalid channel ID"}}
+
+      # Missing permissions
+      iex> Lux.Lenses.Discord.ChannelLens.focus(%{guild_id: "123456789"}, %{})
+      {:error, %{type: "discord_api_error", code: 50001, message: "Missing Access"}}
+
+      # Rate limited
+      iex> Lux.Lenses.Discord.ChannelLens.focus(%{channel_id: "123456789"}, %{})
+      {:error, %{type: "rate_limit", retry_after: 5000}}
+  """
+
+  use Lux.Lens,
+    name: "Discord Channel Lens",
+    description: "Reads Discord channel information",
+    url: "https://discord.com/api/v10",  # Base URL, actual endpoints handled in before_focus
+    method: :get,
+    headers: [{"content-type", "application/json"}],
+    auth: %{
+      type: :custom,
+      auth_function: &__MODULE__.add_auth_header/1
+    },
+    schema: %{
+      type: :object,
+      properties: %{
+        guild_id: %{
+          type: :string,
+          description: "The ID of the guild to list channels from",
+          pattern: "^[0-9]{17,20}$"
+        },
+        channel_id: %{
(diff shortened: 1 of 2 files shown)


lux/lib/lux/prisms/discord/channel_prism.ex
```diff
@@ -0,0 +1,470 @@
...
+        {:error, %{type: "preparation_error", message: e.message}}
+    end
+  end
+
+  @doc """
+  Executes the API request asynchronously with timeout handling.
+  """
+  @spec execute_request(request_params()) :: api_response()
+  def execute_request(request) do
+    task = Task.async(fn ->
```
rparcus, line 348: Why do we need to do it async? We are waiting for it immediately after anyways so what is the point?

lux/lib/lux/prisms/discord/channel_prism.ex
```diff
@@ -0,0 +1,470 @@
...
+    method: :get | :post | :put | :patch | :delete,
+    url: String.t(),
+    body: map() | nil
+  }
+
+  @doc """
+  Adds Discord bot token to request headers.
+  """
+  def add_auth_header(prism) do
+    token = Application.fetch_env!(:lux, :discord_token)
```
rparcus, line 313: It seems like this is not configure anywhere (for example inside `runtime.exs`.

lux/lib/lux/lenses/discord/channel_lens.ex
```diff
@@ -0,0 +1,138 @@
...
+  @doc """
+  Prepares the request URL based on the parameters.
+  If channel_id is provided, fetches a specific channel using GET /channels/{channel.id}.
+  Otherwise, lists all channels in the guild using GET /guilds/{guild.id}/channels.
+  """
+  def before_focus(params) do
+    base_url = "https://discord.com/api/v10"
+
+    url = if Map.has_key?(params, :channel_id) do
+      "#{base_url}/channels/#{params.channel_id}"
```
rparcus, line 98: Wouldn't it be confusing to have one lenses loading either one channel or a list of channels? Why not have two separate lenses for this?

lux/lib/lux/lenses/discord/channel_lens.ex
```diff
@@ -0,0 +1,138 @@
...
+    {:error, %{type: "rate_limit", retry_after: retry_after}}
+  end
+
+  @impl true
+  def after_focus(response) do
+    Logger.error("Unexpected Discord API response: #{inspect(response)}")
+    {:error, %{type: "unexpected_response", message: "Received unexpected response format"}}
+  end
+
+  defp lens_url do
```
rparcus, line 135: this seems to not be used anywhere

lux/lib/lux/prisms/discord/channel_prism.ex
```diff
@@ -0,0 +1,470 @@
...
+          description: "Allowed permission bits as string",
+          pattern: "^[0-9]+$"
+        },
+        deny: %{
+          type: :string,
+          description: "Denied permission bits as string",
+          pattern: "^[0-9]+$"
+        }
+      },
+      required: ["action"],
```
rparcus, line 230: With so many actions hiding inside one schema and the tool being called "ChannelPrism" an llm would have a hard time effectively selecting this tool to perform each action. It would make more sense to focus on something like `Discord.Prisms.CreateMessage`, `Discord.Prisms.UpdateMessage`, `Discord.Prisms.DeleteMessage` and specify the required input there.

I actually don't think that OpenAI or any other providers will support conditionals in schemas.

lux/lib/lux/prisms/discord/channel_prism.ex
```diff
@@ -0,0 +1,470 @@
...
+          description: "Voice channel user limit (0 for unlimited, 1-99 for limit)",
+          minimum: 0,
+          maximum: 99
+        },
+        rtc_region: %{
+          type: :string,
+          description: "Voice channel region override (null for automatic)",
+          nullable: true
+        },
+        video_quality_mode: %{
```
rparcus, line 134: I wonder if we should support video here since the framework doesn't have anything to support it.
It would be best to create a very very very dumb down Prism that can do one thing and one thing only. The API is rich, but we are trying to give simple tools to LLMs to perform specific actions. If we try to expose everything through a schema like this, the LLM will just mess up.

lux/lib/lux/prisms/discord/channel_prism.ex
```diff
@@ -0,0 +1,470 @@
...
+      nil ->
+        Task.shutdown(task)
+        {:error, %{type: "timeout", message: "Request timed out"}}
+    end
+  end
+
+  # Private functions
+
+  @spec make_request(request_params()) :: api_response()
+  defp make_request(%{method: method, url: url, body: body}) do
```
rparcus, line 368: Since we are probably going to do a lot of requests like this, you might want to have a simple Discord Client module. Then you can test that separately from the prisms, and re-use it in each prism.

lux/lib/lux/prisms/discord/channel_prism.ex
```diff
@@ -0,0 +1,470 @@
...
+        {:error, %{
+          type: "unknown_error",
+          message: "An unexpected error occurred",
+          status: response.status
+        }}
+    end
+  end
+
+  @spec format_response(map() | {:error, map()}) :: {:ok, map()} | {:error, map()}
+  defp format_response({:error, _} = error), do: error
```
rparcus, line 414: Formatting also sounds like something that you might want to standardize and reuse as you build more prisms.

rparcus (changes requested):
There is a lot of custom logic here but no tests. If lenses are simply using basic built in functions, then we don't really need to test much, but the prism here was full of new code that should have been tests.

PR #175, opened 2025-03-25
✨Add Discord API client

sojinmm:
Key Features:
- Basic HTTP client with built-in authentication
- Automatic retry mechanism for failed requests
- Rate limit handling with configurable delays
- Comprehensive test coverage using Req.Test

Implementation Details:
- Uses Req for HTTP requests with proper test isolation
- Supports common HTTP methods (GET, POST, PATCH, DELETE)
- Properly handles Discord's rate limits and error responses
- Configurable retry mechanism with default settings
- Clean separation of concerns with private helper functions

Testing:
- Full test coverage for all HTTP methods
- Tests for rate limit handling
- Tests for error responses
- Uses Req.Test for HTTP request mocking
- Proper test isolation and cleanup

### DIFF
--- a/lux/config/config.exs
+++ b/lux/config/config.exs
@@ -6,6 +6,10 @@ config :lux, :open_ai_models,
   default: "gpt-4o-mini",
   smartest: "gpt-4o"
 
+# API Keys
+config :lux, :api_keys,
+  discord: System.get_env("DISCORD_BOT_TOKEN", "test_bot_token")
+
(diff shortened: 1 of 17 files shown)


sojinmm:
Oh, it looks like the previous code got uploaded as well. Since I just need to modify it to use the client, please check the client and client_test for now. I will update Prism to use the client by tomorrow.🥲

PR #178, opened 2025-03-27
✨discord delete message prism

sojinmm:
# DeleteMessagePrism Implementation

DeleteMessagePrism Implementation Features:
- Uses only essential parameters (channel_id, message_id)
- Passes Discord API errors directly for LLM interpretation
- Provides clean success/failure response structure
- Simplified interface by removing optional reason parameter

Test Code Characteristics:
- Uses UnitAPICase for Discord API mocking
- Tests core message deletion functionality
- Verifies HTTP method (DELETE), URL construction, and authorization headers
- Validates success/failure response handling
- Includes schema validation tests

Key Improvements:
- Removed unnecessary reason parameter
- Streamlined error handling with direct Discord API responses
- Added comprehensive documentation
- Enhanced test case clarity

The implemented code features:
- Clean and minimal interface
- LLM-friendly structure for easy understanding and usage
- Comprehensive test coverage
- Clear documentation

### DIFF
--- a/lux/lib/lux/prisms/discord/message/delete_message_prism.ex
+++ b/lux/lib/lux/prisms/discord/message/delete_message_prism.ex
@@ -1,0 +1,94 @@
+defmodule Lux.Prisms.Discord.Messages.DeleteMessagePrism do
+  @moduledoc """
+  A prism for deleting Discord messages.
+
+  This prism provides a simple interface to delete messages from Discord channels. It leverages
+  the Discord API client for making requests and follows a minimalist approach by:
+
+  - Supporting only essential parameters (channel_id, message_id)
+  - Passing through Discord API errors directly for LLM interpretation
+  - Providing clear success/failure responses
+
+  ## Implementation Details
+
+  - Uses Discord API endpoint: DELETE /channels/{channel_id}/messages/{message_id}
+  - Returns a simple success response with the deletion status and message details
+  - Preserves original Discord API errors for better error handling by LLMs
+
+  ## Examples
+
+      # Delete a message
+      iex> DeleteMessagePrism.handler(%{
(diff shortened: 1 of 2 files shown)


lux/lib/lux/prisms/discord/message/delete_message_prism.ex
```diff
@@ -0,0 +1,94 @@
...
+
+  @doc """
+  Handles the request to delete a message from a Discord channel.
+
+  This implementation:
+  - Makes a direct request to Discord API using the Client module
+  - Returns success/failure responses without additional error transformation
+  - Logs the operation for monitoring purposes
+  """
+  def handler(%{channel_id: channel_id, message_id: message_id}, %{agent: agent} = _ctx) do
```
rparcus, line 84: The seconds argument is usually the agent itself (not a map with an agent inside), Also, it would be better if the handler didn't fail because of some context information

example here:

```elixir
def handler(params, agent) do
    # Access agent configuration using Access protocol
    agent_name = agent[:name]

    # Perform health check
    with {:ok, metrics} <- check_system_health(params) do
      Logger.info("Health check completed for #{agent_name}")
      {:ok, metrics}
    end
  end
  ```
  
  For the params, it is ok to map explictly, because those are required params. One issue though is that this will raise if the LLM fails to call the tool with the right args, killing the process. It would be better to validate inputs and return a meaningful error to the LLM instead so it can try again.

rparcus (changes requested):
While testing this I noticed a few bugs in the framework^^

Please help to resolve the comments I left and it will be ready to merge.

image

Logs for using AgentHub and the local router to trigger the `signal_handler`.

image

Added logs to see the LLM picking the right tool (notice the name)

image

PR #181, opened 2025-03-27
✨discord read message lens

sojinmm:
## ReadMessageLens Module
A lens designed to interact with Discord's message reading functionality. It provides a clean interface to fetch messages from Discord channels while handling API responses and errors gracefully.

Key Features:
- Direct Discord API error propagation
- Streamlined response structure
- Built-in authentication handling

The lens transforms raw Discord API responses into a more manageable format, focusing on essential message data like content and author information. When successful, it returns a structured map containing the message content and author details. In case of errors, it propagates Discord's error messages directly to maintain transparency in error handling.

## ReadMessageLens Tests
The test suite ensures the reliability and correctness of the Discord message reading functionality through three main test cases:

1. Message Reading Success
   - Validates proper API endpoint construction
   - Verifies authentication header presence
   - Confirms correct response structure transformation
   - Checks author information mapping

2. Empty Message Handling
   - Ensures proper handling of messages with no content
   - Validates response structure consistency
   - Verifies author data preservation

3. Error Response Handling
   - Tests proper processing of API error responses
   - Validates error message formatting
   - Ensures consistent error state handling

The tests use stubbed HTTP responses to simulate various Discord API scenarios, ensuring the lens behaves predictably across different response types and edge cases.

### DIFF
--- a/lux/lib/lux/integrations/discord/common.ex
+++ b/lux/lib/lux/integrations/discord/common.ex
@@ -1,0 +1,44 @@
+defmodule Lux.Integrations.Discord.Common do
+  @moduledoc """
+  Common settings and functions for Discord API integration.
+  """
+
+  @doc """
+  Common request settings for Discord API calls.
+  """
+  def request_settings do
+    %{
+      headers: [{"Content-Type", "application/json"}],
+      auth: %{
+        type: :custom,
+        auth_function: &__MODULE__.add_auth_header/1
+      }
+    }
+  end
+
+  @doc """
+  Common headers for Discord API calls.
+  """
+  def headers, do: [{"Content-Type", "application/json"}]
+
+  @doc """
+  Common auth settings for Discord API calls.
+  """
+  def auth, do: %{
+    type: :custom,
+    auth_function: &__MODULE__.add_auth_header/1
+  }
+
+  @doc """
+  Adds Discord bot token authorization header.
+  """
+  def add_auth_header(%Lux.Lens{} = lens) do
+    token = Application.get_env(:lux, :api_keys)[:discord]
+    %{lens | headers: lens.headers ++ [{"Authorization", "Bot #{token}"}]}
+  end
+
+  def add_auth_header(conn) do
+    token = Application.get_env(:lux, :api_keys)[:discord]
+    Plug.Conn.put_req_header(conn, "authorization", "Bot #{token}")
+  end
+end
--- a/lux/lib/lux/lenses/discord/messages/read_message.ex
+++ b/lux/lib/lux/lenses/discord/messages/read_message.ex
@@ -45,8 +45,6 @@ defmodule Lux.Lenses.Discord.Messages.ReadMessage do
       required: ["channel_id", "message_id"]
     }
 
-  def before_focus(params), do: params
-
   @doc """
   Transforms the Discord API response into a simpler format.
   ## Examples
@@ -72,9 +70,5 @@ defmodule Lux.Lenses.Discord.Messages.ReadMessage do
 
   def after_focus(%{"message" => message}) do
     {:error, %{"message" => message}}
-  end
-
-  def after_focus(_) do
-    {:error, "invalid_response"}
   end
 end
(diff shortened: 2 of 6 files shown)


lux/lib/lux/lenses/discord/messages/read_message_lens.ex
```diff
@@ -0,0 +1,93 @@
...
+
+  use Lux.Lens,
+    name: "Read Discord Message",
+    description: "Reads a message from a Discord channel",
+    url: "https://discord.com/api/v10/channels/:channel_id/messages/:message_id",
+    method: :get,
+    headers: [{"Content-Type", "application/json"}],
+    auth: %{
+      type: :custom,
+      auth_function: &__MODULE__.add_auth_header/1
```
rparcus, line 30: Can we move this to a common module for all discord Lenses?

lux/test/unit/lux/lenses/discord/messages/read_message_lens_test.exs
```diff
@@ -0,0 +1,110 @@
...
+  """
+
+  use UnitAPICase, async: true
+  alias Lux.Lenses.Discord.Messages.ReadMessageLens
+
+  @channel_id "123456789012345678"
+  @message_id "987654321098765432"
+
+  setup do
+    Req.Test.verify_on_exit!()
```
rparcus, line 17: No point using `verify_on_exit!` if we do not use `expect`. (PS: we should use `expect`)

lux/lib/lux/integrations/discord/common.ex
```diff
@@ -0,0 +1,44 @@
+defmodule Lux.Integrations.Discord.Common do
```
rparcus, line 1: How about this?
```suggestion
defmodule Lux.Integrations.Discord do
```

Then we could just use `Discord.auth()` and similar things. I think it looks a bit cleaner.

lux/lib/lux/integrations/discord/common.ex
```diff
@@ -0,0 +1,44 @@
...
+
+  @doc """
+  Adds Discord bot token authorization header.
+  """
+  def add_auth_header(%Lux.Lens{} = lens) do
+    token = Application.get_env(:lux, :api_keys)[:discord]
+    %{lens | headers: lens.headers ++ [{"Authorization", "Bot #{token}"}]}
+  end
+
+  def add_auth_header(conn) do
```
rparcus, line 40: Let's pattern match on the struct here. I think it's %Plug.Conn{}. Also, since we reuse these everywhere, could we also have some basic `typespecs`?

rparcus (changes requested):
Left a minor comment. We're almost there!

sojinmm:
I’ve made a new commit reflecting the feedback. Do you think it would be a good idea to rename "lens" to something like "prism"? For example, changing "readmessagelens" to "readmessage".

rparcus:
> I’ve made a new commit reflecting the feedback. Do you think it would be a good idea to rename "lens" to something like "prism"? For example, changing "readmessagelens" to "readmessage".

Yes. As mentioned in some other PR, we don't need to keep repeating Lens/Prism in the module name. I think that those words also confuse the LLM, so it would be better to remove them

lux/lib/lux/lenses/discord/messages/read_message.ex
```diff
@@ -0,0 +1,80 @@
...
+        username: author["username"]
+      }
+    }}
+  end
+
+  def after_focus(%{"message" => message}) do
+    {:error, %{"message" => message}}
+  end
+
+  def after_focus(_) do
```
rparcus, line 77: Do we expect anything to come this way? If not, this clase is hiding the error and the developer will have to add logs to see what is actually being returned. Why don't we just remove this clause and let it crash if we receive something unexpected?

lux/lib/lux/lenses/discord/messages/read_message.ex
```diff
@@ -0,0 +1,80 @@
...
+        message_id: %{
+          type: :string,
+          description: "The ID of the message to read",
+          pattern: "^[0-9]{17,20}$"
+        }
+      },
+      required: ["channel_id", "message_id"]
+    }
+
+  def before_focus(params), do: params
```
rparcus, line 48: Hmm is it mandatory to implement this function even if we don't need it to change anything? If yes, then I need to change something in `Lux.Lens`, but if not, please remove it.

rparcus (approved):

PR #182, opened 2025-03-27
Goat sdk uniswap swap

coffee-converter:


lux/priv/python/pyproject.toml
```diff
@@ -7,14 +7,17 @@ readme = "README.md"
 packages = [{include = "lux"}]
 
 [tool.poetry.dependencies]
-python = "^3.12.3"
-web3 = "^6.15.11"
-eth-typing = "3.5.2"
+python = "^3.10"
```
rparcus, line 10: woooo do we have to do this?

lux/test/unit/lux/prisms/goat_sdk/uniswap_swap_test.exs
```diff
@@ -0,0 +1,124 @@
+defmodule Lux.Prisms.GoatSdk.UniswapSwapTest do
+  use UnitCase, async: true
+
+  alias Lux.Prisms.GoatSdk.UniswapSwap
+
+  @moduletag :unit
+
+  setup do
+    mock_python_code = """
```
rparcus, line 9: `setup do` blocks run before each `test`. Was this the goal here? If you need this code to exist in the system we could add a python file inside `lux/privy/python/goat`

lux/lib/lux/prisms/goat_sdk/uniswap_swap.ex
```diff
@@ -0,0 +1,120 @@
+# lib/lux/prisms/goat_sdk/uniswap_swap.ex
```
rparcus, line 1:
```suggestion
```

lux/lib/lux/prisms/goat_sdk/uniswap_swap.ex
```diff
@@ -0,0 +1,120 @@
...
+defmodule Lux.Prisms.GoatSdk.UniswapSwap do
```
rparcus, line 2: This looks great actually and thanks for the effort of testing the python part. I'm thinking of how to make the process a little easier.

Not sure if you'd be willing to re-write this, but Lux also supports [pure python prisms](https://github.com/sojinmm/lux/blob/da30e4b5c72b54e68eead9260248bc4984be6002/lux/guides/prisms.livemd#L540-L541). We then could have simpler pure python tests, I think.

All the python action happens in the `priv` folder.

lux/priv/python/pyproject.toml
rparcus: If we change the python version, then we should also update a bunch of other places such as `.tool-version` and CI... a lot of CI...

lux/lib/lux/prisms/goat_sdk/uniswap_swap.ex
```diff
@@ -27,6 +27,11 @@ defmodule Lux.Prisms.GoatSdk.UniswapSwap do
           type: :integer,
           description: "Slippage tolerance in basis points (e.g. 50 for 0.5%)",
           default: 50
+        },
+        api_key: %{
```
rparcus, line 31: We shouldn't pass API keys as args. We should load them from config. Please take a look at `config/runtime.exs` and `Lux.Config` to see how we expose Configs to the whole app.

I am thinking now that we don't have a native way of doing it for python code (I actually don't know what's the usual way to do it^^)

lux/lib/lux/prisms/goat_sdk/uniswap_swap.ex
```diff
@@ -85,27 +95,66 @@ defmodule Lux.Prisms.GoatSdk.UniswapSwap do
                to_token: params.to_token,
                amount: params.amount,
                chain_id: params.chain_id,
-               slippage: params.slippage
+               slippage: params.slippage,
+               api_key: params.api_key
```
rparcus, line 99: As mentioned above, should be loaded from configs, not passed as an arg.

lux/priv/python/pyproject.toml
coffee-converter: I can try without this downgrade...  I think cursor suggested this in trying to get goat-sdk to work

PR #206, opened 2025-03-31
Fix typos in rule file.

chitacan:
Found some typos in rule file.


rparcus (approved):
😅 Thanks!

(I should stop manually typing these prompts...)

PR #250, opened 2025-04-07
Update agent configuration form

chitacan:
Update form component for agent configuration in lux_app.

## Changes

- Added fields for agent's llm configuration (provider, model, temperature)
- Added lux components (Beams, Prisms, Lenses) selector to the form
- Change export format to compatible with lux's agent loader

CleanShot 2025-04-07 at 21 13 42@2x

* #221
  * #249 👈

### DIFF
--- a/lux_app/assets/js/consts.js
+++ b/lux_app/assets/js/consts.js
@@ -1,0 +1,6 @@
+export const NODE_WIDTH = 210;
+export const NODE_HEIGHT = 110;
+export const NODE_GAP_X = 20;
+export const NODE_GAP_Y = 20;
+export const NODE_RADIUS = 10;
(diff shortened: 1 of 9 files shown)


chitacan:
@rparcus In this PR, I implemented components (Beam, Lens, Prism) selector in the agent's form.
I think this UI is much more straightforward and intuitive when adding lux components to an agent.
The existing edge connection UI seems to be suitable for generating signals between agents. What do you think about this?

Adding lux components to an agent:

https://github.com/user-attachments/assets/0d22b34b-0872-40a1-8bda-cfbe1ad91d2a

Adding signals between agents (needs more work): 

https://github.com/user-attachments/assets/6558fcad-c484-4ff5-9a9c-ffa904cf61ba

Issue #54, opened 2025-02-25
Discord Signals System ($500)


rparcus:
# Core Signal Types
Implement the signal schemas to handle discord communication.

Examples:
- DiscordMessageSignal
    - Content
    - Attachments
    - Embeds
    - Components
    - Metadata (channel, guild, author)
- DiscordInteractionSignal
    - Slash commands
    - Buttons
    - Select menus
    - Modals
    - Context menus
- DiscordPresenceSignal
    - Member status
    - Activities
    - Guild presence
etc..

**Acceptance Criteria**:
- Complete signal type definitions
- Bidirectional conversion (Discord ↔ Signals)
- Signal validation and schema
- Signal processing pipeline
- Error handling for each signal
- Tests for all signal types

sojinmm:
@rparcus I can take this up?

rparcus:
Hey there, would you like on the whole discord feature then? For this one, how would you like to start? Are you already familiar with Signals, Prisms, Lenses, etc?

sojinmm:
> Hey there, would you like on the whole discord feature then? For this one, how would you like to start? Are you already familiar with Signals, Prisms, Lenses, etc?

Hi! Yes, I'll be working on the Discord feature implementation. I've been reviewing the existing codebase and Discord API documentation.

I see the basic patterns for Signals, Prisms, and Lenses in our current implementation, and I plan to follow those patterns while implementing the Discord features. I'm thinking of starting with the Channel structure and its related Signal, following our existing schema patterns.

Would you agree with this approach? And do you have any specific considerations or requirements for the Discord integration that I should keep in mind?

rparcus:
Hi Sojin, this is cool. I think we can assign this task first and work on the scope as we go. 

Any particular reason to start with Channels? How about we try to think of a specific "user story" such as "An agent is able to join a server" , "An agent is able to send a message to a server", "and agent is able to read mentions in a server", "an agent is able to read all message in a channel".

I think it would be great to have an idea about what capabilities you would want to enable for agents, and then we can think together about the building blocks necessary to achieve that.

PS: "Channels" seem certainly something we should have a representation for.

sojinmm:
Yes, in that case, I think it would be good to start with core functionalities like "an agent is able to send a message to a server" and "an agent is able to read mentions in a server." Should I push small PRs directly here as I go?

> Hi Sojin, this is cool. I think we can assign this task first and work on the scope as we go.
> 
[...]

PR #458, opened 2026-03-08
feat: Discord Signals System (#54)

Susuhome:
Discord signal schemas and processing pipeline for issue #54.

## Signal Types
- **MessageSignal** — content, attachments, embeds, components, metadata
- **InteractionSignal** — slash commands, buttons, select menus, modals, context menus
- **PresenceSignal** — member status, activities, guild presence

## Features
- Bidirectional conversion (Discord ↔ Signal structs)
- Signal validation
- Classification (auto-detect signal type from payload)
- Processing pipeline (composable steps, halt on error)

## Tests: 20 tests, 0 failures
Budget: $500

### DIFF
--- a/lux/lib/lux/signals/discord/interaction_signal.ex
+++ b/lux/lib/lux/signals/discord/interaction_signal.ex
@@ -1,0 +1,61 @@
+defmodule Lux.Signals.Discord.InteractionSignal do
+  @moduledoc "Discord interaction signal: slash commands, buttons, select menus, modals."
+
+  defstruct [:type, :name, :data, :user, :channel_id, :guild_id, :token, :metadata, :timestamp]
+
+  @interaction_types [:slash_command, :button, :select_menu, :modal, :context_menu]
+
+  def new(attrs) when is_map(attrs) do
+    type = attrs[:type] || attrs["type"]
+    if type in @interaction_types do
+      {:ok, %__MODULE__{
+        type: type,
+        name: attrs[:name] || attrs["name"],
+        data: attrs[:data] || attrs["data"] || %{},
+        user: attrs[:user] || attrs["user"],
+        channel_id: attrs[:channel_id],
+        guild_id: attrs[:guild_id],
+        token: attrs[:token],
+        metadata: attrs[:metadata],
+        timestamp: DateTime.utc_now()
+      }}
+    else
+      {:error, :invalid_interaction_type}
+    end
+  end
+
+  def to_discord(%__MODULE__{} = signal) do
+    %{
+      "type" => type_to_int(signal.type),
+      "data" => %{"name" => signal.name, "options" => signal.data},
+      "channel_id" => signal.channel_id,
+      "guild_id" => signal.guild_id
+    }
+  end
+
+  def from_discord(payload) when is_map(payload) do
+    new(%{
+      type: int_to_type(payload["type"]),
+      name: get_in(payload, ["data", "name"]),
+      data: get_in(payload, ["data", "options"]) || %{},
+      user: payload["member"] || payload["user"],
+      channel_id: payload["channel_id"],
+      guild_id: payload["guild_id"],
+      token: payload["token"]
+    })
+  end
+
+  def interaction_types, do: @interaction_types
+
+  defp type_to_int(:slash_command), do: 2
+  defp type_to_int(:button), do: 3
+  defp type_to_int(:select_menu), do: 3
+  defp type_to_int(:modal), do: 5
+  defp type_to_int(:context_menu), do: 2
+  defp type_to_int(_), do: 0
+
+  defp int_to_type(2), do: :slash_command
+  defp int_to_type(3), do: :button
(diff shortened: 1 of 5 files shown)


MyTH-zyxeon:
Review-assist pass for #54 / PR #458:

This looks like a useful starting point for the Discord signal schema layer. A few acceptance/merge risks stood out that I would tighten before maintainer review:

1. Preserve the Discord interaction subtype. Right now `InteractionSignal.from_discord/1` maps raw type `3` to `:button`, so select menus and other message components collapse into the same signal. For Discord, `type=3` means message component and the real subtype is in `data.component_type`; context menu commands also need the application-command subtype preserved instead of round-tripping as `:slash_command`.

2. Make incoming payload parsing safe for unknown external values. `PresenceSignal.from_discord/1` uses `String.to_existing_atom/1`, which can raise on unexpected status strings. A bounded parser returning `{:error, :invalid_status}` or a safe fallback would fit the rest of the API better.

3. Normalize string-key support consistently. `MessageSignal.new/1` accepts both atom and string keys, but `InteractionSignal.new/1` and `PresenceSignal.new/1` still read several fields only from atom keys (`channel_id`, `guild_id`, `token`, `user_id`, etc.). Real Discord gateway/API payloads will be string-keyed, so these fields can silently disappear in round trips.

4. Prove integration with the existing Discord primitives. The PR adds standalone signal structs and a pipeline, but I do not see tests that connect them to the existing Discord lenses/prisms/beams from the earlier Discord work. A small test that converts an actual message/interaction/presence payload into a signal and then into the downstream operation/input shape would make the bounty acceptance much easier to judge.

5. Add error-path and edge-case tests. Good follow-ups: unknown interaction type, unknown `data.component_type`, missing/empty required Discord IDs, presence status not in the allowlist, atom-key and string-key parity for every schema, and pipeline halting while preserving the original error reason.

No secrets or live Discord API calls should be needed for those checks; mocked representative Discord payload fixtures should be enough.

MyTH-zyxeon:
Follow-up review-assist pass for #54 / PR #458 after rechecking the current head (`73c84ae`). I am not claiming the bounty here; just adding a tighter acceptance checklist for the existing implementation.

Good small patch targets before maintainer review:

1. Strengthen the bidirectional conversion tests beyond the current content/channel assertions. `MessageSignal.to_discord/1` drops `author`, `guild_id`, `attachments`, `metadata`, and timestamp information, while `from_discord/1` accepts some of those fields. Some of that may be intentionally one-way for outbound Discord payloads, but the bounty asks for `Discord <-> Signals`; a regression test should either preserve signal metadata on round trip or document which Discord-only fields are intentionally not emitted.

2. Preserve interaction response context on round trip. `InteractionSignal.from_discord/1` captures `user`, `channel_id`, `guild_id`, and `token`, but `to_discord/1` only emits top-level type/data/channel/guild. If downstream pipeline steps need to answer an interaction, losing `token`/user context can make the signal valid but unusable for follow-up responses.

3. Align `context_menu` modeling with Discord's application command shape. `type_to_int(:context_menu)` currently emits top-level type `2`, same as slash commands, but context menus are differentiated by command metadata under `data.type`/command target type. A test with a representative user/message context-menu payload would catch this.

4. Make validation cover schema-required identifiers, not just non-empty message content. Examples worth testing: message without `channel_id`, interaction without `token`, presence without `user_id`, and malformed activity maps. That would make the "Signal validation and schema" acceptance criterion easier to verify.

5. Tighten the pipeline contract. The `@type step` says a step receives/returns maps, but `process/2` passes `%MessageSignal{}` / `%InteractionSignal{}` / `%PresenceSignal{}` structs in the tests. Either widen the type/spec to signal structs or normalize parsed signals into maps before running pipeline steps.

All of those can stay fully mocked with fixture payloads; no Discord bot token or live API call is needed.

MyTH-zyxeon:
Follow-up review-assist pass for #54 / PR #458 at head `73c84aea181c553f6741791e3ab73b89ec88d7fe`.

Two remaining acceptance gaps I would tighten before maintainer review:

1. The new Discord modules are plain structs (`MessageSignal`, `InteractionSignal`, `PresenceSignal`) and do not define `Lux.SignalSchema` modules or wrap/validate through `Lux.Signal` with `schema_id`. Since #54 asks for Discord signal schemas plus validation/processing, it would be stronger to show how a Discord message/interaction/presence becomes a first-class Lux signal envelope that existing agent signal handling can route and validate, not only a standalone struct consumed by `Lux.Signals.Discord.Pipeline`.

2. `from_discord/1` currently discards source timestamps and `new/1` stamps every parsed payload with `DateTime.utc_now()`. For Discord events, that makes replay, ordering, duplicate detection, and audit trails harder: the signal timestamp becomes parse time instead of event time. Please consider preserving Discord message timestamps / presence event timestamps in `timestamp` or `metadata`, with tests that prove parsed historical fixtures keep their original event time.

No Discord token, bot session, live gateway, or API call was used for this static review.

PR #536, opened 2026-05-13
feat: add DeFi analytics integration

### LINKED ISSUE #88: DeFi Analytics Integration $750
## Overview

Implement DeFi analytics integration with DeFiLlama and Dune Analytics for comprehensive protocol analysis.

## Core Features
- TVL tracking
- Protocol metrics
- Yield analytics
- Volume analysis
- Custom query support
- Historical data access
- Analytics dashboard
- Alert system

## Acceptance Criteria
- [ ] TVL data integration
- [ ] Protocol metrics collection
- [ ] Yield tracking system
- [ ] Volume analytics implementation
- [ ] Query system integration
- [ ] Integration tests for analytics
- [ ] Dashboard implementation

## Budget: \$750

BigPolarBear:
## Summary
- Add `Lux.Integrations.DeFiAnalytics` settings plus a Req-based client for DeFiLlama TVL/yields/DEX volume APIs and Dune query execution/result APIs.
- Add deterministic analytics helpers for protocol metrics, TVL history, yield ranking, DEX volume analytics, custom Dune rows, dashboard data, and threshold alerts.
- Add Livebook guide plus unit/integration coverage for client endpoints, payload normalization, dashboard, and alerts.

## Tests
- `mix test test/unit/lux/integrations/defi_analytics --include unit`
- `mix test test/integration/defi_analytics_workflow_test.exs --include integration`
- `mix test.unit`
- `git diff --cached --check`

Wallet for bounty payout: `0x38938CB6Fd54172C5E82a5d01a9236fF12E84BdF`

Fixes #88


MyTH-zyxeon:
Thanks for the implementation. I reviewed this against the #88 acceptance criteria and found a few gaps that would be worth tightening before merge:

- `Analytics.dashboard/3` appears to only work correctly with raw DeFiLlama protocol rows. If callers pass the output of `collect_protocol_metrics/1` or `process_protocols/1`, the second normalization pass looks for `tvl`, `change_1d`, and `change_7d` again and can zero out already-normalized `tvl_usd` / `change_*_pct` values. A regression test for dashboard input produced by the collector would catch this.
- Custom Dune query support is implemented at the client/result-normalization layer, but the dashboard path does not consume custom query rows or expose them in the returned analytics payload. That leaves the "custom query support" acceptance criterion disconnected from the dashboard implementation.
- Dune execution is asynchronous in practice, but this PR does not show a polling/timeout path that ties `dune_execute_sql/2`, `dune_execution_status/2`, and `dune_execution_result/2` into one safe workflow. Without that, callers have to assemble execution lifecycle behavior themselves.
- Alerts are returned as computed data, but there is no alert delivery/scheduling boundary or documented contract for how an agent/UI should consume, persist, de-duplicate, or acknowledge alert events.
- The tests cover deterministic fixtures well, but I do not see failure-path coverage for DeFiLlama/Dune rate limits, malformed rows, missing API keys, or Dune execution errors. Those cases are important for analytics jobs that run unattended.

Suggested next step: add one end-to-end "collector -> dashboard -> alerts" test using normalized collector output, then wire Dune custom query rows into the dashboard payload with an explicit execution lifecycle helper and fake-client tests for success, timeout, and error states.

MyTH-zyxeon:
Thanks, I rechecked PR #536 at head `c46e2028e4daefcdf6c3b096b71e5bc99ef4addc`. A few additional implementation-level gaps stood out beyond my earlier acceptance-criteria pass:

- `pct_change/3` compares against `Enum.at(values, -periods - 1)`, so a 7-day window with exactly 7 non-zero samples returns `nil` and reports `0.0`. That can hide real TVL/volume movement; either require 8 samples explicitly or compare first vs last for the available 7-sample window.
- `normalize_series/3` only handles map rows, while the adjacent `normalize_pair_series/3` already supports `[timestamp, amount]` and `{timestamp, amount}` chart points. If DeFiLlama TVL-like series arrive in pair form, the TVL path zeroes them out; TVL normalization should get the same pair support or a fixture proving the API shape is always map-only here.
- Dune API key handling is inconsistent across the guide and runtime: the Livebook guide reads `DUNE_API_KEY`, the integration reads `Application.get_env(:lux, :api_keys)[:dune]`, and `Client.headers/2` silently sends no auth header when neither is present. A single documented config contract plus an early `{:error, :missing_api_key}` path would make unattended analytics jobs safer.
- `request/4` builds per-call options first, then merges `Application.get_env(:lux, Client, [])` afterward. That lets global app env override explicit per-call `plug`, `headers`, `params`, `json`, or `retry`, which can make tests and one-off jobs surprising. Defaults should merge first, with explicit call options winning last.
- Yield alert ranking currently only discounts small TVL and positive `ilRisk`. I would add fixtures for stale data, missing TVL, negative APY, reward-heavy APY, stable/volatile pool type, and chain allowlist behavior so high-yield dashboard alerts do not over-promote unsafe or stale pools.

Suggested next slice: add focused regression tests for the 7-sample `pct_change/3` case and pair-form TVL series, then tighten Dune config/early-error behavior before expanding dashboard alert risk scoring.

MyTH-zyxeon:
Quick follow-up at head `c46e2028e4daefcdf6c3b096b71e5bc99ef4addc`: since there do not appear to be commits after the May 30 recheck, here is a compact merge-readiness checklist for the #88 DeFi analytics acceptance criteria.

- **Dashboard input contract:** `Analytics.dashboard/3` should accept either raw DeFiLlama rows or already-normalized outputs from `collect_protocol_metrics/1`, `process_protocols/1`, `collect_volume_metrics/1`, and `process_volume_overview/1` without re-zeroing `tvl_usd`, `change_*_pct`, or `total_*_usd`. Please add regression tests for both raw and normalized collector output.
- **Historical series support:** make TVL normalization handle map rows and pair rows consistently, matching `normalize_pair_series/3`, or add a fixture proving the DeFiLlama TVL endpoint used here never returns `[timestamp, value]` pairs. The current map-only path makes historical TVL analytics fragile.
- **Dune lifecycle helper:** wire custom query support into one safe helper such as `run_dune_query/2` that executes, polls status with timeout/backoff, fetches results, and returns typed errors. That would make the custom query acceptance criterion usable from agents instead of leaving callers to assemble the lifecycle.
- **Config precedence:** merge app defaults first and explicit per-call options last, so tests/jobs can override `plug`, `headers`, `params`, `json`, `retry`, `base_url`, and `api_key` predictably. Missing Dune auth should return a clear `{:error, :missing_api_key}` before building a request.
- **Alert contract:** define dashboard alert shape and lifecycle fields such as `id`, `source`, `metric`, `threshold`, `observed_at`, `dedupe_key`, and `ack_state`, then add tests for persistence-ready output rather than only computed in-memory alerts.
- **Risk fixtures:** add deterministic yield/alert fixtures for stale data, missing TVL, negative APY, reward-heavy APY, stable vs volatile pools, and chain allowlist behavior so dashboards do not promote unsafe high-yield pools.

That would turn the implementation into a safer review slice: collector-compatible dashboards, explicit Dune query lifecycle behavior, deterministic alert semantics, and failure-path coverage for unattended analytics jobs.

PR #559, opened 2026-05-14
feat: add Rust component definitions

### LINKED ISSUE #103: Rust Component Definition Support $500
## Overview

Implement support for defining Lux components (Prisms and Beams) in Rust, enabling high-performance native components with full framework integration.

## Core Features
- Rust component definition
- Trait system integration
- Async/await support
- Performance optimizations
- Component lifecycle management

## Technical Details
- Implement Rust component definition system
- Create trait system integration
- Set up async/await support
- Add performance optimizations
- Implement component lifecycle management

## Acceptance Criteria
- [ ] Implement Rust component definition system
- [ ] Create trait system integration
- [ ] Set up async/await support
- [ ] Add performance optimizations
- [ ] Implement component lifecycle management
- [ ] Add documentation for component definition
- [ ] Include examples of Rust components

## Budget: $500

BigPolarBear:
Closes #103.

## Summary
- Adds `Lux.Rust` for Rust component definitions, a `~RS` sigil, Prism/Beam/Lens metadata builders, Cargo manifest generation, dependency/cache plans, lifecycle plans, async runtime plans, performance metadata, validation, and Rust trait templates.
- Adds a checked `priv/rust` crate with `LuxComponent`, `LuxPrism`, `LuxBeam`, `LuxLens`, lifecycle states, error types, and an async handler helper.
- Adds a Rust validation-prism example and Rust language-support guide registered in ExDoc extras.

## Verification
```bash
mix test.unit test/unit/lux/rust_test.exs
cargo check --manifest-path priv/rust/Cargo.toml
mix credo --strict lib/lux/rust.ex test/unit/lux/rust_test.exs
MIX_ENV=prod mix compile
git diff --cached --check
```

All passed locally. Production compile still shows pre-existing Allora/Mira warnings unrelated to this change.


MyTH-zyxeon:
Review-assist pass for the $500 Rust Component Definition Support bounty in #103.

This is a useful documentation/metadata foundation, but I see a scope risk against the issue's "full framework integration" acceptance criteria:

1. The current `Lux.Rust` module says it does not compile or load NIFs by itself, and the PR does not appear to wire a runtime adapter into Lux prisms/beams. That makes the implementation more of a definition/planning layer than an executable Rust component path. For #103, I would expect at least one end-to-end path where a Rust-defined Prism or Beam can be registered/invoked from Lux, or a clearly scoped follow-up if this PR intentionally stops before execution.

2. Async support is mostly metadata right now. `runtime_plan/1` records Tokio/backpressure settings, and `handle_async/2` in `priv/rust/src/lib.rs` wraps the synchronous `handle/1` trait method, but there is not an async trait contract or bridge showing an actual awaited Rust component operation. A small async component example/test would make that criterion much stronger.

3. Lifecycle management is represented as a plan and enum, but I do not see compile/load/start/call/stop state transitions being enforced or observed. Tests currently validate map shape rather than the lifecycle behavior. Adding a unit test or adapter test that exercises lifecycle transitions would reduce maintainer ambiguity.

4. The generated `Cargo.toml` helper and the checked-in `priv/rust/Cargo.toml` can drift. It would help to add a regression test that generated manifest output remains aligned with the checked-in Rust crate, or explicitly document which one is canonical.

5. The example component is good, but the Elixir tests do not seem to compile/check the Rust example. If CI cannot run Cargo everywhere, a documented command plus a focused optional CI/test target would still make the acceptance evidence clearer.

In short: this PR is a good scaffold, but I would not call it complete for #103 until the execution bridge, async contract, and lifecycle behavior are either implemented or explicitly split into follow-up issues.

MyTH-zyxeon:
Follow-up review assist for #103 / PR #559 after checking the current diff again:

The earlier comment already covered the main scope question around execution, async, and lifecycle behavior. A few narrower implementation/test gaps also look worth closing before this is treated as complete for the Rust component-definition bounty:

- `component_template/1` sets metadata through `traits_for(kind)` as if generated Prisms/Beams implement `LuxPrism` / `LuxBeam`, but the emitted Rust template only implements `LuxComponent`. The checked-in `validation_prism.rs` adds `impl LuxPrism for ValidationPrism {}`, so the generated template and the example are inconsistent. A template regression test for Prism and Beam output would catch this.
- `cargo_manifest/1` accepts dependency maps, but `render_dependencies/1` only reads atom keys with `Map.fetch!(opts, :version)` and `Map.get(opts, :features, [])`. If component metadata is loaded from JSON/YAML or agent-generated maps with `"version"` / `"features"` string keys, manifest rendering will crash even though other helpers in this module intentionally accept both atom and string keys through `value/3`.
- The `rust do ... end` macro path is not exercised by tests, and its `case block do` has no fallback branch. A non-sigil block or a slightly different quoted shape will raise a compile-time `CaseClauseError` instead of a useful error. Either testing the documented macro form or removing the unused macro surface would reduce support risk.
- `component_definition/1` returns `{:ok, definition}` for blank `name`, `module`, or `handler` values because `required/2` only rejects `nil`; the separate validator catches blanks later. If downstream code assumes `{:ok, definition}` means invokable metadata, this can leak invalid components into runtime planning.
- The generated dependency/build plan points at the shared `priv/rust/Cargo.toml`, but there is no per-component source/manifest materialization step in the plan. If the intended output is only a reusable scaffold, the docs should say that; if the plan is meant to compile a generated component, it needs a concrete write/build/load handoff.

I still did not run Cargo, compile native code, or execute the branch locally; this is a static review pass intended to make the maintainer acceptance checks sharper.

MyTH-zyxeon:
Rechecked #103 / PR #559 at head `0c07f8f807c5f10eb773f0a20263c719994ab872`. Beyond the earlier execution-bridge and template/dependency notes, I think there is one more merge-readiness area to tighten: the generated Rust/Cargo surface needs stricter validation before user or agent-provided metadata reaches generated code.

Concrete gaps:

- `component_template/1` normalizes `kind` but does not reject unknown kinds the way `component_definition/1` does. `kind: :worker` would emit `ComponentKind::Worker`, which is not a Rust enum variant, while `kind: "worker"` would hit `Atom.to_string/1` and crash. A template-level guard/test would catch this before invalid Rust is produced.
- `validate_component/1` only checks that `traits` is non-empty; it does not verify that `kind: :prism` includes `:lux_prism`, `kind: :beam` includes `:lux_beam`, etc. That means a definition can validate as a Rust component while advertising the wrong Lux trait integration.
- `component_template/1` interpolates `:struct` and `:name` directly into Rust code. Values with spaces, punctuation, reserved Rust keywords, or generated names from external metadata can produce invalid code or surprising tokens. This should probably either sanitize to a Rust identifier or return a structured validation error.
- `cargo_manifest/1` interpolates `package`, `version`, and `edition` into TOML strings without escaping or format validation. If those values are loaded from JSON/YAML, a quote/newline in the package metadata can generate malformed Cargo.toml. Even a small validator for crate/package names and escaped TOML strings would make the build plan safer.
- The tests currently prove happy-path map shapes and one Beam template, but not these failure boundaries. I would add tests for invalid template kind, string-keyed/agent-provided metadata, trait/kind mismatch, invalid Rust struct identifiers, and package names that must be rejected or escaped.

That would make the definition layer a safer foundation even if runtime execution remains a later adapter, because generated Rust and Cargo files would fail fast at the Lux boundary instead of at Cargo compile time.

PR #563, opened 2026-05-14
feat: add multi-chain aggregation core

### LINKED ISSUE #74: Multi-Chain Data Aggregation Engine $1,750
## Overview

Develop a comprehensive multi-chain data aggregation system for collecting and processing blockchain data across multiple EVM-compatible networks.

## Core Features
- Multi-chain RPC management
- Block and transaction monitoring
- Smart contract event aggregation
- Historical data indexing
- Real-time data streaming
- Data normalization across chains
- Configurable data retention
- Query optimization

## Acceptance Criteria
- [ ] Support for major EVM chains (Ethereum, BSC, Polygon, etc.)
- [ ] Real-time block and transaction monitoring
- [ ] Smart contract event filtering and processing
- [ ] Efficient data storage and retrieval system
- [ ] Query interface for aggregated data
- [ ] Rate limiting and error handling
- [ ] Integration tests demonstrating multi-chain data collection

## Budget: \$1,750

BigPolarBear:
## Summary
- Adds a transport-neutral multi-chain EVM aggregation core for RPC request planning, response normalization, querying, retention policy, health, and latency summaries.
- Adds Lux prisms for building request plans, aggregating collected responses, and querying datasets.
- Covers default major EVM chains and explicit chain metadata with RPC URLs, rate-limit metadata, log filters, latest block reads, balances, and transaction count plans.

Closes #74

## Verification
- mix format lib/lux/integrations/web3/multi_chain_aggregation.ex lib/lux/prisms/web3/multi_chain/build_request_plan.ex lib/lux/prisms/web3/multi_chain/aggregate_responses.ex lib/lux/prisms/web3/multi_chain/query_dataset.ex test/unit/lux/integrations/web3/multi_chain_aggregation_test.exs test/unit/lux/prisms/web3/multi_chain/build_request_plan_test.exs test/unit/lux/prisms/web3/multi_chain/aggregate_responses_test.exs test/unit/lux/prisms/web3/multi_chain/query_dataset_test.exs
- git diff --check
- MIX_ENV=prod mix compile
- Core smoke: {2, %{ethereum: 16}, 1, "0xabc"}
- Prism smoke: {1, %{ethereum: 2}, 1}

## Known test blocker
- mix test.unit test/unit/lux/integrations/web3/multi_chain_aggregation_test.exs test/unit/lux/prisms/web3/multi_chain is blocked before these tests run by existing Credo compile failure on Elixir 1.19.5: Regex.CompileError in credo/check/config_comment_finder.ex:9.


MyTH-zyxeon:
Review assist for #74 acceptance criteria:

This is a clean transport-neutral core for building RPC request plans and aggregating already-collected responses. Before counting it as the full Multi-Chain Data Aggregation Engine bounty, I would close a few acceptance-criteria gaps:

1. The issue asks for multi-chain data collection, real-time block/transaction monitoring, and event processing. This PR builds request payloads and normalizes caller-provided responses, but it does not include an RPC executor, polling/streaming loop, subscription boundary, or fake-client integration test that proves blocks/logs are actually collected across chains.

2. Rate limiting and retries are described as metadata only. `plan_requests/2` returns `summary.rate_limit`, and each request has a static `retry` map, but no code enforces concurrency, retry attempts, backoff, or rate-limit failures. A small executor behaviour with deterministic tests would make this acceptance criterion concrete without needing live RPC credentials.

3. Historical indexing and retention are policy-only. `retention_plan/1` describes compaction, but the implementation keeps everything in-memory and does not expose a storage/index boundary or a pruning/query test that proves old blocks/logs/transactions can be retained and retrieved efficiently.

4. `transaction_count` and `balance` actions can emit JSON-RPC params with a nil address when the filter has no address. That should probably return a validation error, or the prism should document that those actions require an address.

5. Tests are unit-level around normalized maps. The bounty asks for integration tests demonstrating multi-chain data collection; a mocked Ethereum+Polygon RPC client exercising latest block + logs + query would cover that safely and avoid any live network dependency.

With those pieces, this would be much closer to the full collection/monitoring/indexing engine described in #74 rather than just the planning/aggregation layer.

MyTH-zyxeon:
Follow-up pass focused on a different merge-risk area than the earlier executor/rate-limit/retention notes:

1. The prism handlers fetch required inputs with string-key support, but then use `Map.take(params, [:actions, :filter, :request_id_start, :rate_limit])` and `Map.take(params, [:generated_at, :retention])`. If the prism receives normal JSON-decoded params with string keys, `BuildRequestPlan` silently drops caller-supplied actions, filters, request ids, and rate-limit config; `AggregateResponses` similarly drops `generated_at` and retention overrides. `QueryDataset` avoids this because it passes the string-key map through to `normalize_options/1`. I would add string-key handler tests for all three prisms and normalize options before `Map.take`.

2. `query/2` passes `limit` and `offset` straight into `Enum.take/2` and `Enum.drop/2` after integer parsing. Negative values are accepted by those APIs and change the query semantics by taking/dropping from the tail. Since this is exposed as a dataset-query prism, invalid negative pagination should probably return a validation error or clamp to safe non-negative values, with tests for `limit: -1` and `offset: -1`.

3. Log rows normalize addresses, but topics are stored as provided and `filter_topic/2` does a literal membership check. That means a caller can query the same event topic with different hex casing and miss rows even though addresses are case-normalized. Lowercasing/normalizing topics in both ingestion and query would make the aggregation behavior consistent.

4. `:transactions` rows are only `Map.put(&1, :chain, chain)`, so transaction hashes, from/to addresses, block numbers, and values are not normalized the way log rows are. If transaction monitoring is part of the bounty acceptance path, a small transaction fixture test should prove normalized address/hash/block fields and queryability across chains.

These are all no-network fixes, so they should be testable without live RPC credentials while making the prism-facing API much safer for real agent use.

MyTH-zyxeon:
Follow-up review for #74 / PR #563 after re-checking the current diff against the bounty acceptance criteria.

This PR is a solid transport-neutral planning and normalization layer, but I still see a few merge/acceptance risks that are separate from the earlier string-key and rate-limit notes:

1. The issue asks for real-time block/transaction monitoring and multi-chain data collection, but this PR still stops at building request plans and aggregating caller-provided responses. There is no polling/scheduling loop, RPC executor, backfill cursor, subscription/websocket path, or retry queue that actually collects blocks/logs/transactions across chains.

2. Historical indexing and retention are modeled as a policy, but not enforced. `retention_plan/1` returns compaction metadata, yet there is no persisted store, partition key, pruning job, replay cursor, or queryable index boundary. That makes the "efficient data storage and retrieval" and "configurable data retention" criteria hard to validate from this PR alone.

3. Balance and transaction-count actions can generate invalid RPC params when the filter does not include an address. `build_rpc_request/4` for `:transaction_count` and `:balance` uses `List.first(filter.address)` and can produce `[nil, block]`, which will become a remote RPC error instead of a local validation error.

4. `QueryDataset` can receive JSON-decoded datasets with string keys through the prism boundary, but `query/2` reads atom keys like `:logs`, `:transactions`, `:chain`, `:address`, and `:topics`. Aggregated in-process datasets work, while externally supplied JSON-shaped datasets can silently return zero rows.

5. The current tests exercise in-memory helpers only. For #74 acceptance, I would add at least one integration-style fixture that runs a multi-chain request plan through mocked Ethereum/Polygon/BSC RPC responses into aggregation and query, including an error/rate-limit response and a historical block range.

No RPC endpoint, private key, wallet, API token, live chain request, or funds-moving workflow was used for this review.

Issue #96, opened 2025-02-25
Local Model Support via Ollama $400


rparcus:
## Overview

Implement local model support via Ollama, enabling self-hosted LLM capabilities with optimized performance.

## Core Features
- Ollama integration
- Local model management
- Model download and caching
- Performance optimization
- Resource management

## Technical Details
- Implement Ollama client
- Add model management system
- Set up caching mechanisms
- Implement resource controls
- Add performance monitoring

## Acceptance Criteria
- [ ] Implement Ollama API client
- [ ] Add model management functionality
- [ ] Create caching system for models
- [ ] Implement resource usage controls
- [ ] Add performance monitoring
- [ ] Include documentation and examples
- [ ] Add integration tests

## Budget: $400

BHARATJHAWAR52:
Hi @rparcus 
I am interested in contributing to this project where we can run local models via Ollama.
I have an experience in building models, and analyzing data, for various purposes. Also, I know about integrating Ollama with different projects.

1914Jegx:
/attempt #96

Implementation plan:
1. Implement Lux.LLM.Ollama stack (client, resource pool, model manager, supervisor) with telemetry and configuration hooks.
2. Wire runtime/application config and env vars so Ollama defaults work locally and in tests.
3. Add documentation plus unit/integration coverage validating model caching, auto-pull, and request flow.

IshGadol:
Hi team,

I’d like to take on this bounty and implement Ollama support for Lux. My plan is straightforward:

- Build a dedicated Ollama provider that plugs into Lux’s existing LLM integration layer.
- Add an HTTP client for Ollama’s API (/api/chat, /api/generate, /api/list, /api/pull, /api/show, /api/delete, etc.).
- Provide model management helpers (list, pull, show, delete) with light caching so models are pulled on-demand and tracked.
- Map Lux runtime options to Ollama’s resource controls (num_ctx, temperature, top_p, keep_alive, etc.).
- Support both streaming and non-streaming responses, plus JSON/Schema outputs using Ollama’s format parameter.
- Emit performance telemetry (load duration, tokens/sec, eval times) so metrics can feed into Lux’s monitoring.
- Ship a minimal working example (Beam) and documentation showing setup, config (OLLAMA_BASE_URL, OLLAMA_MODEL, etc.), and usage.
- Add tests: unit tests with mocked responses, and optional integration tests gated on a local Ollama install.

A couple of quick clarifications to align with your expectations:
[...]

PR #564, opened 2026-05-14
feat: add Ollama local model support

BigPolarBear:
## Summary
- Adds Lux.LLM.Ollama for local Ollama chat completions via /api/chat.
- Adds local model management helpers for list/show/pull/delete/ensure workflows.
- Adds TTL model-list caching, resource-control planning, and performance metric normalization for Ollama duration/token fields.
- Adds mocked unit coverage for chat payloads, model cache/pull/delete paths, resource warnings, and tool conversion.

Closes #96

## Verification
- mix format lib/lux/llm/ollama.ex test/unit/lux/llm/ollama_test.exs
- git diff --check
- MIX_ENV=prod mix compile
- Core smoke: {"llama3.2", 65536, 2, 10.0, 2000.0}

## Known test blocker
- mix test.unit test/unit/lux/llm/ollama_test.exs is blocked before these tests run by the existing Credo compile failure on Elixir 1.19.5: Regex.CompileError in credo/check/config_comment_finder.ex:9.


MyTH-zyxeon:
Exact-head review assist for `faeb81b08766cc7b58b265452137a185c350b83e` against the open #96 ($400) bounty. The model-management and metrics helpers are useful, but three execution blockers remain:

1. **A model response can execute a tool that was never offered to it.** `call/3` sends the requested `tools`, but `handle_chat_response/1` discards that allowlist and passes every returned function name to `execute_tool/2` (`ollama.ex:85-103, 386-463`). That code reconstructs an arbitrary module name, loads it, and invokes any Lux Prism or Beam it finds. A hallucinated or malicious local model response can therefore invoke capabilities outside the caller's tool set. Carry the offered tool-name-to-module map into response handling, reject unknown names before module resolution, and add a test where the response names a valid Lux tool that was not supplied.

2. **Parallel tool-call failures can perform later side effects and then crash.** `execute_tool_calls/1` eagerly runs every call through `Enum.map/2`, then reduces the results (`ollama.ex:429-435`). For `[{:error, reason}, {:ok, result}]`, the second reducer step has no clause for an `{:error, reason}` accumulator and raises `FunctionClauseError`; meanwhile the later tool has already run. Process calls with an explicit policy (fail-fast or collect all results), preserve order, and test an error followed by a successful side-effecting tool.

3. **The public `stream: true` path cannot consume Ollama's streamed response.** The config exposes `stream`, sends it unchanged to `/api/chat`, and always handles the 200 body as one final JSON object (`ollama.ex:57, 77, 94, 101-103, 386-413`). Ollama streams partial chat fields that must be accumulated, especially `tool_calls`; there is no chunk/NDJSON accumulation or streaming test here. Either force/reject streaming until it is implemented, or accumulate chunks and cover partial content/tool calls plus mid-stream errors. Official behavior: https://docs.ollama.com/capabilities/streaming and https://docs.ollama.com/capabilities/tool-calling.

GitHub exposes zero exact-head check-runs. The only submitted tests are tagged `:unit`, while the repository test helper excludes unit and integration tests by default, and #96 explicitly requires integration coverage. Please provide a command/result that actually includes this suite and a separately gated local-Ollama integration path before closing the bounty.

Run-Id: `run-5857f83f-d544-4136-a5c0-b858ed1d32ca`
Trace-Id: `f921925a-f03e-4c29-a2a2-51f8d8c0ae22`
Requester: Jun via Codex desktop automation `monitor-sxt-bounty-pr-1757`

PR #569, opened 2026-05-14
feat: add Rust testing framework integration

### LINKED ISSUE #102: Rust Testing Framework Integration $350
## Overview

Implement a testing framework for Rust code in Lux, enabling seamless integration with the existing test infrastructure and comprehensive test coverage.

## Core Features
- Rust test runner
- Integration with mix test
- Coverage reporting
- Cross-language test utilities
- Test fixtures and helpers

## Technical Details
- Implement Rust test runner
- Create integration with mix test
- Set up coverage reporting
- Add cross-language test utilities
- Implement test fixtures and helpers

## Acceptance Criteria
- [ ] Implement Rust test runner
- [ ] Create integration with mix test
- [ ] Set up coverage reporting
- [ ] Add cross-language test utilities
- [ ] Implement test fixtures and helpers
- [ ] Add documentation for testing
- [ ] Include examples of Rust tests

## Budget: $350

BigPolarBear:
Fixes #102.

## Summary
- Adds `Lux.Rust` facade plus `Lux.Rust.TestRunner` for cargo test planning, execution, output summaries, ExUnit after-suite hook, and LCOV parsing.
- Adds `Lux.Rust.Fixtures` for Elixir/Rust fixture manifests and generated fixture projects.
- Adds `mix rust.test` / `mix test.rust`, defaulting to `:prod` so Rust tests do not compile dev/test-only deps.
- Adds a minimal `priv/rust` example crate with unit + integration tests.
- Documents Rust testing setup, coverage env, fixtures, and optional `mix test` hook.

## Verification
- `git diff --check`
- `MIX_ENV=prod mix compile`
- `cargo test --manifest-path priv/rust/Cargo.toml`
- `mix rust.test`

Blocked locally:
- `mix test.unit test/unit/lux/rust/test_runner_test.exs test/unit/lux/rust/fixtures_test.exs test/unit/mix/tasks/rust_test_test.exs`
- Existing blocker: `:credo` fails to compile under Elixir 1.19.5 with `Regex.CompileError` in `lib/credo/check/config_comment_finder.ex:9` before these tests run.

Payout wallet: `0x38938CB6Fd54172C5E82a5d01a9236fF12E84BdF`.


MyTH-zyxeon:
Review-assist pass for the $350 Rust Testing Framework Integration bounty in #102.

This PR is the stronger open implementation path I found for #102, but I see one blocker-shaped issue and a couple of acceptance-clarity gaps:

1. `mix.exs` appears to add `"rust.test": :prod` inside the `aliases` list while also defining `Mix.Tasks.Rust.Test`. That can shadow the real `mix rust.test` task with an alias value that is not the task implementation. I think the env selection belongs only under `cli/0` `preferred_envs`, while `"test.rust": "rust.test"` can stay as the convenience alias. Please verify `mix rust.test --project-root priv/rust` dispatches to `Mix.Tasks.Rust.Test.run/1` after this change.

2. The task module lives at `lib/mix/tasks/rust/test.ex`, so the canonical task name should already be `rust.test`. A regression test that invokes `Mix.Task.run("rust.test", ...)` without relying on the alias table would protect this integration point.

3. The coverage path sets LLVM env vars and parses LCOV, but I do not see a test around `--coverage` task invocation proving the env is passed to the runner/system opts. A focused test there would make the "coverage integration" acceptance evidence much clearer.

4. The fixtures bridge is useful, but the current Rust test only reads `LUX_RUST_FIXTURE_LABEL` from the environment. If the intended cross-language fixture contract is manifest/file based, it would be stronger to include one test that reads the generated `lux_rust_fixtures.json` or a fixture file path, not only an environment variable.

The biggest thing to check is item 1, because it may make the advertised `mix rust.test` command fail or call the wrong thing even though the implementation module itself looks useful.

MyTH-zyxeon:
Follow-up review assist for #102 / PR #569 after rechecking the current diff:

The previous review already called out the task dispatch, direct `Mix.Task.run("rust.test", ...)` coverage, coverage-env testing, and fixture-manifest bridge. A few narrower acceptance gaps are still worth tightening:

- `mix rust.test --coverage --coverage-output cover/rust` currently appears to set LLVM coverage env vars and return a coverage plan in the summary, but it does not run a coverage tool or write/parse an LCOV report as part of the task. Since #102 explicitly asks for coverage reporting, either the task should produce a report path (for example by invoking the configured tool after `cargo test`) or the docs should state that `--coverage` only instruments the run and LCOV generation is a separate manual step.
- `--coverage-tool` is parsed and converted with `String.to_atom/1`, but I do not see the selected tool used anywhere beyond metadata. If the CLI accepts arbitrary tool names, it would be safer to keep this as a bounded string or validate it against known atoms like `:llvm_cov` / `:grcov`.
- `install_ex_unit_hook/1` is the only path that claims `mix test` integration, but the tests do not exercise the hook behavior. A small test with an injected runner, or a doc note that this is an opt-in after-suite gate rather than automatic `mix test` integration, would make that acceptance criterion clearer.
- Fixture manifests include `generated_at` by defaulting to `DateTime.utc_now/0`. That is useful provenance, but it means repeated runs over identical fixtures do not produce byte-identical manifests unless the caller injects a timestamp. If "deterministic fixtures" is a goal, consider separating deterministic file hashes from volatile run metadata or documenting the timestamp override.

I still did not run Cargo, compile native code, or execute this branch locally; this is a static pass focused on making the maintainer's acceptance checks sharper without requiring live services or secrets.

MyTH-zyxeon:
Follow-up review assist for #102 / PR #569 at head `be77b54d9a7fb32181b8b8ee554e73115bb7c5a7`.

I rechecked the current `mix.exs` and the earlier task-dispatch concern looks addressed now: I see `test.rust` as the convenience alias, with `rust.test` handled by `Mix.Tasks.Rust.Test` and env prefs under `preferred_cli_env` / `cli/0`.

Two remaining integration details I would tighten before merge:

1. `Lux.Rust.project_path/0` expands `priv/rust` from `File.cwd!/0`. That is fine for `mix rust.test` when run from the Lux repo root, but the public facade examples (`Lux.Rust.build_test_plan/2`, `Lux.Rust.run_tests/2`) can point at the caller's current working directory if Lux is used as a dependency or called from another app. For a library-facing default, consider resolving from `:code.priv_dir(:lux)` when available, with the cwd-based path only as a dev fallback.

2. `Lux.Rust.Fixtures.write_manifest!/3` calls `Jason.encode!/2`, but `mix.exs` still does not declare `:jason` as a direct dependency even though this module now calls it directly. Jason is present transitively today through existing deps, but making it direct (or switching to Elixir's built-in JSON support if that is the intended minimum-version path) would make the package contract less fragile.

I still did not run this branch locally or execute Cargo/native code; this is a static acceptance pass focused on merge-readiness for the Rust testing bounty.

PR #589, opened 2026-05-14
ci: cache Python dependency downloads

### LINKED ISSUE #31: Cache python modules in CI

popsiclelmlm:
## Summary
- add Poetry installer caching to Python security scan and dependency update workflows
- cache pip and Poetry package download directories with the Python lockfile as the cache key
- clean trailing whitespace at the end of the touched workflow files

## Verification
- git diff --check
- ruby -e 'require "yaml"; ARGV.each { |f| YAML.load_file(f); puts "ok #{f}" }' .github/workflows/security-scan.yml .github/workflows/dependency-updates.yml

Closes #31


MyTH-zyxeon:
Review-assist pass for #31 / this PR. I did not run workflow jobs or use any secrets; this is a static review of the CI cache changes.

This is pointed in the right direction, but I found one correctness risk that should probably be fixed before merge:

1. The cache key uses `hashFiles('lux/priv/python/poetry.lock')`, but the default branch currently does not expose `lux/priv/python/poetry.lock` via the repository contents API. The Python tree has `pyproject.toml` but no lockfile. That means the cache key can collapse to the same empty-hash value across dependency changes, so stale Poetry/pip artifacts may be restored even when Python dependencies change. If this repo intentionally does not commit a lockfile, keying on `lux/priv/python/pyproject.toml` (or adding/committing the lockfile if that is the desired workflow) would make the cache invalidation deterministic.

Two smaller hardening points:

2. `~/.local` is broad for the Poetry installer cache. A narrower Poetry path, or including Poetry version plus Python version in the key, would reduce accidental reuse if the workflow later installs additional tools into `~/.local`.

3. Because `virtualenvs-in-project` is false, the useful Poetry virtualenv/cache state is under Poetry's cache directories. It would be good to state whether the intent is only download-cache speedup or full virtualenv reuse, then keep the cached paths aligned with that intent.

With the missing-lockfile key fixed, this should be a safer CI cache improvement rather than a cache that silently stops invalidating.

PR #592, opened 2026-05-14
feat: Telegram Analytics & Monitoring

### LINKED ISSUE #67: Telegram Analytics and Monitoring ($1,200)
## Overview
Implement prisms and lenses for tracking and analyzing Telegram bot usage and performance.

## Core Features
- Message analytics
- User engagement tracking
- Command usage statistics
- Error rate monitoring
- Performance metrics
- Usage pattern analysis
- Custom metric definitions

## Acceptance Criteria
- [ ] Analytics data collection system
- [ ] Metric aggregation prisms
- [ ] Performance monitoring tools
- [ ] Custom reporting capabilities
- [ ] Test coverage > 90%
- [ ] Data visualization helpers

## Budget: $1,200

zp6:
Closes #67

Complete Telegram analytics & monitoring system ($1,200 bounty):

- Analytics engine: message tracking, user activity, sentiment analysis
- Monitor: real-time alerting, keyword detection, health checks
- Supervisor: OTP supervision tree
- 6 prisms: generate_report, get_stats, track_metric
- Full test suite (12 test files)
- 1690 lines following Lux patterns (moduledoc, @spec, {:ok, _} | {:error, _})

GitHub: @zp6

### DIFF
--- a/lux/lib/lux/integrations/telegram/analytics.ex
+++ b/lux/lib/lux/integrations/telegram/analytics.ex
@@ -1,0 +1,533 @@
+defmodule Lux.Integrations.Telegram.Analytics do
+  @moduledoc """
+  Analytics engine for tracking and analyzing Telegram bot usage and performance.
+
+  Provides an ETS-backed metrics collection system that supports:
+
+  - **Message analytics** — track sent/received message counts, throughput
+  - **User engagement** — active users, new users, retention windows
+  - **Command usage** — per-command invocation counts and success rates
+  - **Error monitoring** — error counts, rates, and last error details
+  - **Performance metrics** — response-time histograms and percentiles
+  - **Custom metrics** — arbitrary key-value counters and gauges
+  - **Usage patterns** — hourly/daily activity breakdowns
+
+  All data is stored in an ETS table (`:telegram_analytics`) managed by this
+  GenServer, making it fast for reads while keeping writes lightweight.
+
+  ## Architecture
+
+  The analytics engine is started under `Lux.Integrations.Telegram.Supervisor`
+  and registers itself as `__MODULE__` so other processes can call it directly.
+
+  Metrics are grouped into *namespaces* (`:messages`, `:users`, `:commands`,
+  `:errors`, `:performance`, `:custom`) to keep the key-space organised.
+
+  ## Quick Start
+
+      # Track a sent message
+      Analytics.track(:messages, :sent, 1)
+
+      # Track a command invocation
+      Analytics.track(:commands, "/start", 1, meta: %{user_id: 123})
+
+      # Record a response time
+      Analytics.record_response_time(150)
+
+      # Get aggregated stats
+      {:ok, stats} = Analytics.get_stats(:messages)
+  """
+
+  use GenServer
+
+  require Logger
+
+  @table :telegram_analytics
+  @namespaces [:messages, :users, :commands, :errors, :performance, :custom]
+
+  # ── Types ──────────────────────────────────────────────────────────────
+
+  @type namespace :: :messages | :users | :commands | :errors | :performance | :custom
+  @type metric_key :: atom() | String.t()
+  @type metric_value :: number() | map()
+  @type time_window :: :hour | :day | :week | :all
+
+  @type stats :: %{
+          total: non_neg_integer(),
+          metrics: %{metric_key() => metric_value()},
+          window: time_window(),
+          computed_at: DateTime.t()
+        }
+
+  # ── Client API ─────────────────────────────────────────────────────────
+
+  @doc """
+  Starts the analytics engine and creates the ETS table.
+  """
+  @spec start_link(keyword()) :: GenServer.on_start()
+  def start_link(opts \\ []) do
+    name = Keyword.get(opts, :name, __MODULE__)
+    GenServer.start_link(__MODULE__, opts, name: name)
+  end
+
+  @doc """
+  Tracks a metric increment under the given namespace.
+
+  ## Parameters
+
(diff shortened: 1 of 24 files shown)


MyTH-zyxeon:
Review assist for #67, focused on the listed Telegram analytics and monitoring acceptance criteria:

1. `Monitor.track_call/2` and `TrackMetric.handler/2` build atoms from runtime strings with `String.to_atom/1` (`"endpoint_#{endpoint}"` and arbitrary metric keys). If endpoint names or custom keys can come from bot traffic or agent inputs, that can leak atoms over time. Tuple/string metric keys, a bounded registry, or `String.to_existing_atom/1` for known keys would make the monitoring path safer.

2. The bounty asks for data visualization helpers. The current implementation returns structured report maps and hourly buckets, but I did not see chart/table helper payloads, visualization-ready series builders, or examples showing how reports feed a dashboard. Adding a small `to_series`/`to_chart_data` helper for usage, error, and latency data would close that criterion more directly.

3. Several tests only assert that async calls do not crash (`assert true`) after sending metrics. It would be stronger to assert the stored counter/gauge/error/latency outputs through the public stats/report APIs, including an integration path from `Monitor.track_call/2` into `Analytics.get_stats/1` and `GenerateReport`.

4. The analytics table is process-local ETS with hourly buckets in GenServer state. That may be fine if in-memory monitoring is intended, but the issue mentions analytics data collection and custom reporting. A short retention/export/import story, or explicit docs that this is ephemeral runtime telemetry, would prevent reviewers from expecting persistent analytics.

5. The PR includes a large Discord prism surface in addition to the Telegram analytics work. If that scope is unrelated to #67, splitting it out would reduce review risk and make it easier to validate the Telegram bounty criteria on their own.

6. The report supports `:hour`, `:day`, `:week`, and `:all`, but the current `compute_namespace_stats/1` always returns `window: :all`; window filtering mostly applies to usage patterns. If the intent is full-window reporting, the namespace stats should also respect the selected window or the report should label that distinction clearly.

MyTH-zyxeon:
Follow-up review-assist pass for the `$1,200` Telegram Analytics and Monitoring bounty (#67), focused on implementation/test correctness beyond the earlier atom-safety, visualization, persistence, and scope notes.

The current PR has a useful analytics/monitoring shape, but I would tighten these before maintainers depend on it:

1. The public `Analytics.*` API always talks to the globally registered `Lux.Integrations.Telegram.Analytics` name, while tests start the GenServer with unique names such as `:"analytics_test_#{...}"`. That means several tests are not exercising the process they started, and `GenServer.call(__MODULE__, ...)` paths can fail when only the named test process exists. Either pass a server/name through the API or start the global process in tests.

2. A few tests use `send(pid, {:track, ...})`, but the implementation handles `{:track, ...}` through `handle_cast/2`, not `handle_info/2`. Those messages will not execute the metric update path. The tests should call `Analytics.track/4` or `GenServer.cast/2` against the intended server and then assert real stats.

3. `build_report/2` looks up performance data with `:ets.lookup(@table, {{:performance, :response_time}})`, but `record_response_time/1` stores the key as `{:performance, :response_time}`. Because of the extra tuple wrapper in the lookup, reports can show zero/default performance even after latency metrics were recorded.

4. `Monitor.track_call/2` runs the supplied Telegram API function inside the monitor GenServer call with an infinite timeout. A slow or stuck Telegram request can block health checks, alert registration, and other monitoring calls. It would be safer to record timing around the call outside the GenServer, or isolate the monitored call in a task with a bounded timeout.

5. `Monitor.track_call/2` increments `:messages/:sent` for every successful wrapped endpoint, including non-message endpoints like `getMe` or admin/list operations. That can inflate message analytics and error-rate denominators. Endpoint classification or explicit metric type would make the analytics more accurate.

Suggested verification slice: start analytics under a non-global test name and prove the public API can target it, assert `record_response_time/1` appears in `generate_report/1`, exercise a slow tracked function without blocking `get_status/0`, and verify `getMe` does not count as a sent message.

MyTH-zyxeon:
Follow-up2 review assist for the `$1,200` Telegram Analytics and Monitoring bounty (#67 / #592), after rechecking `6bb021b2540bf231c5fa6156bf95d39f3b17ec4b`. I do not see a newer commit after the previous correctness pass, so this is a separate acceptance-risk checklist:

1. `Analytics.start_link/1` always creates the named ETS table `:telegram_analytics`, even when the GenServer itself is started with a unique `:name`. Starting a second analytics process in the same VM can fail with `:badarg` because the named table already exists. That is especially relevant because the tests use unique process names; a restart/parallel test path can still collide on the global ETS table. Either derive the table from the server instance, accept a configurable table name, or document that only one analytics engine may exist.

2. Several prism handlers read only atom keys from `params` (`params[:namespace]`, `params[:key]`, `Map.get(params, :window, "day")`). If the runtime passes JSON-shaped string keys from the input schema, `GetStats` and `TrackMetric` can see `nil` values or default to `day` despite a provided `"window"`. Supporting both atom and string keys would make the prism boundary much safer.

3. Response-time alerts only run on the error path. `Monitor.check_alerts/4` can trigger `:response_time`, but it is called only when `track_call/2` returns `{:error, reason}`. A slow successful Telegram API call records latency but never checks registered response-time alerts. A regression test with a successful slow function and `metric: :response_time` would make this visible.

4. `record_user_event/2` accepts any atom as the event and stores the whole event list per user indefinitely. For long-running bots, that can grow without retention or sampling and make the advertised retention analytics ambiguous. If the product goal is active/new/retained users, storing bounded counters or timestamped last-seen windows may be safer than an unbounded event history list.

These are separate from the earlier atom-safety, visualization, persistence, global-name, and performance lookup notes. They would help maintainers verify that the analytics subsystem behaves predictably in normal bot runtimes and agent/prism calls.

Issue #68, opened 2025-02-25
YouTube Core Integration and Live Streaming ($3,000)


rparcus:
## Overview
Implement core YouTube platform integration with advanced live streaming capabilities, enabling Lux agents to manage YouTube channels programmatically.

## Core Features
- OAuth2 authentication and token management
- Video upload and management
- Live stream scheduling and management
- Stream health monitoring
- Real-time chat interaction
- Broadcast parameter management
- Multi-stream quality control
- Automated stream transitions (testing → live)

## Acceptance Criteria
- [ ] Complete YouTube Data API v3 integration
- [ ] Live streaming setup and management
- [ ] Real-time chat monitoring and interaction
- [ ] Stream health monitoring system
- [ ] Documentation and examples

## Budget: $3,000

mudaly10-coder:
Auto-claim: taking this on (HAF).
Will submit a minimal PR quickly; happy to align scope/acceptance if needed.

Treasure520520:
/attempt #68

I am working on a focused YouTube Core + Live Streaming slice for this bounty: OAuth2 PKCE helpers, authenticated Data API client with retry/rate-limit handling, live broadcast/stream/chat helpers, docs, and focused tests. I will submit a PR shortly and keep the implementation aligned with existing Lux integration patterns.

rafael81:
/attempt #68

Submitted PR #543 for a focused YouTube core integration slice: https://github.com/Spectral-Finance/lux/pull/543

Scope includes OAuth2 consent/token helpers, Data API request normalization, resumable upload initiation, live broadcast/stream helpers, live chat read/write, stream health lookup, docs, and unit tests.

Validation is included in the PR body. This work was prepared with AI assistance and manually reviewed before submission.

rafael81:
/opire try

Submitted PR: https://github.com/Spectral-Finance/lux/pull/543

Validation and AI-assistance disclosure are included in the PR body.

BigPolarBear:
Implemented in PR: https://github.com/Spectral-Finance/lux/pull/545

Scope:
- YouTube OAuth helpers and Data API v3 client coverage for channels, video metadata upload sessions, live broadcasts, live streams, transitions, and live chat
- Live streaming utilities for health monitoring, transition planning, multi-stream quality checks, chat normalization, and dashboard snapshots
- Livebook guide for OAuth, video upload session setup, live setup, chat, health, and transitions

Verification:
[...]

PR #616, opened 2026-05-17
[bounty #68] Add YouTube core live integration

jianmosier:
Addresses #68.

## Summary
- Add a YouTube integration module and client for Data API, upload, and OAuth token endpoints.
- Add OAuth exchange/refresh prisms, resumable upload startup, channel lookup, live broadcast/stream/chat lenses, stream health inspection, and live broadcast transition/bind/chat prisms.
- Add a Livebook guide plus focused unit coverage for the new YouTube modules.

## Validation
- `mix format --check-formatted` on the touched Elixir files
- `MIX_ENV=prod mix compile`
- `mix test --no-deps-check test/unit/lux/integrations/youtube/client_test.exs test/unit/lux/lenses/youtube/channels/list_channels_test.exs test/unit/lux/lenses/youtube/live/list_broadcasts_test.exs test/unit/lux/lenses/youtube/live/list_chat_messages_test.exs test/unit/lux/lenses/youtube/live/stream_health_test.exs test/unit/lux/prisms/youtube/live/live_prisms_test.exs test/unit/lux/prisms/youtube/oauth/oauth_prisms_test.exs test/unit/lux/prisms/youtube/videos/start_resumable_upload_test.exs --include unit` -> 18 tests, 0 failures
- `git diff --check`

Note: full plain `mix test` is blocked on this local machine by the locked `credo 1.7.11` dev/test dependency failing to compile under locally installed Elixir 1.19/OTP 28. The project pins Elixir 1.18/OTP 27 in `.tool-versions`; the targeted YouTube tests above were run after compiling the required runtime/test dependencies without Credo.

## Disclosure
This PR was prepared with AI assistance and manually reviewed/validated before submission.

/claim #68


MyTH-zyxeon:
I noticed one integration blocker that is worth tightening before this is treated as a complete #68 workflow.

The OAuth prisms return an `access_token`, but the downstream YouTube action prisms look for `params[:token]` while their input schemas do not declare either `token` or `access_token`. In a schema-driven Lux invocation, `ExchangeCode` / `RefreshToken` cannot be composed directly into `CreateBroadcast`, `CreateStream`, `SendChatMessage`, `StartResumableUpload`, etc.; those actions will fall back to `Lux.Config.youtube_access_token()` instead of the freshly authorized channel token.

That makes the per-channel OAuth flow and multi-channel live/chat operations fragile: a user can successfully exchange a code, but the follow-up live broadcast/chat calls may run against a static configured token or fail if no global token is present.

Suggested fix:
- Add an optional `access_token` input to the YouTube action prism schemas.
- Thread `params[:access_token] || params[:token]` into the client request.
- Let `Client` accept both `:access_token` and `:token` for consistency with the OAuth prism output.
- Add one test that chains `ExchangeCode` output into a live/chat action and asserts the request uses the exchanged token in the Authorization header.

That would make the OAuth acceptance path much stronger for the bounty because the token exchange and the live-streaming operations become an actual composable workflow instead of separate pieces.

MyTH-zyxeon:
Follow-up after rechecking PR #616 for bounty #68. I am not claiming the bounty here; this is a review-assist pass on the existing submission.

- `config/runtime.exs` now requires `YOUTUBE_ACCESS_TOKEN`, `YOUTUBE_CLIENT_ID`, and `YOUTUBE_CLIENT_SECRET` in dev/test. Unless the normal test/bootstrap path loads `test.envrc`, this can make unrelated local dev or CI test runs fail before they reach the new YouTube tests.
- The OAuth prisms return `access_token` / `refresh_token`, and the low-level client accepts a `token:` option, but the lens modules (`ListChannels`, `ListBroadcasts`, `ListChatMessages`, `StreamHealth`) still authenticate through `YouTube.auth()` and the configured static token. That makes the OAuth flow hard to compose for multi-user/channel agents unless the lenses expose a runtime token or credential context.
- `RefreshToken` returns a new access token but does not surface the refresh token again, expiry handling strategy, or a persistence hook. That leaves long-running live workflows unclear once the access token expires mid-stream.
- `TransitionBroadcast.handler/2` allows any requested target status and does not use `LiveStream.transition_plan/2`. The helper knows about safe `ready -> testing -> live` sequencing, but callers can still request direct transitions that may fail or skip the intended guardrail.
- The live chat path reads messages and sends one message, but it does not demonstrate repeated polling using `next_page_token` / `polling_interval_millis` or moderation/error-rate behavior, which are important for the "live" part of the bounty.
- The upload prism starts a resumable upload session and returns an upload URL, but the PR does not show the next byte-upload step, completion verification, or persisted video ID/status after the session is used.
- Tests are broad and useful, but they mostly verify one mocked success response per prism/lens. I would add explicit tests for missing env bootstrap, dynamic OAuth token composition, token refresh expiry, unsafe live transition requests, paged chat polling, upload-session failure/no-location, and common YouTube 403/quota/rate-limit responses.

A focused next pass would make token context first-class across both lenses and prisms, keep dev/test config from requiring real YouTube env by default, enforce or expose the safe transition plan in the transition prism, and add failure/refresh/pagination tests around the live workflow.

MyTH-zyxeon:
Additional review-assist pass for #68 / PR #616.

I re-read the current diff with the existing OAuth/token comments in mind. A couple of lower-level issues still look worth fixing because they can make the new YouTube tools fail even after the token-threading path is cleaned up:

1. Most prism handlers read required inputs with `Map.fetch(params, :atom_key)` and optional fields with `params[:atom_key]`. If the Lux runtime passes schema-driven JSON-style string keys, values like `"title"`, `"scheduled_start_time"`, `"broadcast_id"`, `"live_chat_id"`, `"refresh_token"`, or `"content_length"` will be treated as missing. The lens `before_focus/1` helpers have the same atom-key-only rename path for `max_results`, `page_token`, `broadcast_status`, etc. A small key-normalization helper shared by lenses/prisms would make these APIs much safer for actual agent/tool invocation.

2. Several lens tests mock `Req.Test.expect(Lux.Lens, ...)`, while the direct client/prism tests mock `YouTubeClientMock`. That may be intentional through the existing `Lux.Lens` HTTP path, but it leaves an important gap: the new `YouTube.auth()` header injection and `Lux.Integrations.YouTube.Client.raw_request/3` path are not tested through the same adapter target. I would add one end-to-end mocked lens test that proves the YouTube auth function, renamed params, and final request all hit the expected Req test plug without falling back to a real network request.

3. The OAuth and live-stream prisms do not validate returned YouTube resource shape before normalizing. For example, `LiveStream.normalize_broadcast/1` and `normalize_stream/1` require an `"id"` key; a successful but partial/malformed mocked response will raise a function-clause error instead of returning `{:error, ...}`. That is especially important for quota/permission edge cases where Google may return unexpected response bodies.

4. `ListChannels.normalize_channel/1` uses `String.to_integer/1` for count fields. YouTube can omit counts such as hidden subscriber counts; if a field is present but non-numeric or empty, this raises instead of returning `nil` or an error tuple. A fixture for hidden/omitted subscriber stats would cover that.

None of this requires a live YouTube account or real OAuth credentials; mocked request/response fixtures should be enough to keep the bounty acceptance path deterministic.

PR #617, opened 2026-05-18
Cache Dialyzer PLTs by runtime version

### LINKED ISSUE #18: Cache dialyzer in CI
Because life is too short

Image

dashan12344423:
Fixes Spectral-Finance/lux#18

This updates CI Dialyzer PLT caching so cache keys include the OTP/Elixir runtime and Mix project inputs, reducing stale or incompatible PLT reuse.

Changes:
- Include OTP/Elixir version, mix.exs, mix.lock, and .tool-versions in Dialyzer cache keys
- Apply the same cache key strategy across Lux and Lux App workflows
- Remove a duplicate Dialyzer cache step in lux-app-ci.yml so restore/build/save behavior is consistent

Validation:
- Workflow-only change; reviewed updated cache keys across .github/workflows


MyTH-zyxeon:
Review assist for #18 / #617.

This is a useful direction for making the PLT cache runtime-aware. One merge-risk I noticed while checking the workflow set:

- `.github/workflows/ci.yml` still has a Lux App "Cache Dialyzer PLTs" step, but that job does not build or run Dialyzer for `lux_app` afterward. So in that workflow the new `lux_app` runtime-aware cache key is still never exercised or refreshed by a `mix dialyzer --plt` / `mix dialyzer` sequence. `lux-app-ci.yml` and `main-ci.yml` have the explicit restore/build/save/run pattern; `ci.yml` looks inconsistent.

A safe follow-up would be either to remove the unused Lux App PLT cache from `ci.yml` if that job is intentionally test-only, or mirror the `main-ci.yml` restore/build/save/run flow there so all workflows that cache PLTs also validate and refresh them.

Small robustness note: the cache key hard-codes `otp-27.2-elixir-1.18.1` in multiple workflow files while the setup versions live in the workflow YAML and `.tool-versions` only records the Elixir/OTP major pairing. Including the workflow file or setup-version source in the key would reduce drift risk when OTP/Elixir versions change later.

PR #632, opened 2026-05-28
feat: add Ollama LLM integration (#96 - local model support)

JirA44:
## Summary
Implements Lux.LLM.Ollama for local model support via Ollama OpenAI-compatible endpoint.

- Follows existing Lux.LLM.OpenAI pattern
- Supports tools (Beams, Prisms, Lenses), JSON mode, configurable timeout
- Handles :econnrefused with clear error message
- Unit tests with Req.Test mocks

Closes #96


MyTH-zyxeon:
Review assist for #96 / PR #632:

This is a concrete Ollama local-model implementation attempt for the #96 bounty: it adds `Lux.LLM.Ollama` plus unit tests using `Req.Test`, without requiring a live Ollama daemon in the test path.

- Ollama integration: the module targets Ollama's OpenAI-compatible `/v1/chat/completions` endpoint and defaults to a local endpoint.
- Local model management path: model selection is configurable through call config and application config.
- Tool support: the implementation converts Beams, Prisms, and Lenses into function specs and handles returned tool calls.
- JSON mode and timeout support: the request body supports JSON-format mode, and the request options include a configurable receive timeout.
- Test coverage: tests mock request/response flow, connection-refused handling, tool calls, JSON format, and function-spec conversion.

Merge gates I would check before accepting:
- Confirm whether the OpenAI-compatible Ollama endpoint expects `max_tokens` rather than `num_predict`; `num_predict` is common in Ollama's native API but may not be honored on `/v1/chat/completions`.
- Confirm the tool-name round trip works for real Elixir module names after replacing `.` with `_` and then rebuilding the module with `Module.concat/1`.
- Confirm `Application.get_env(:lux, Lux.LLM.Ollama, [])` is not merged into the Req request options with provider config keys that Req does not expect.
- Confirm CI runs the new unit file without a local Ollama process and without network access; the mocked `:econnrefused` path should be sufficient.

I did not start Ollama, make any live local-model request, use an API key, or execute model-generated code for this review pass.

MyTH-zyxeon:
Follow-up review-assist pass for #96 / PR #632.

I re-read the current diff after the earlier checklist and found a few more concrete acceptance risks that are worth tightening before maintainer review:

1. The tests appear to register `Req.Test.expect(Ollama, ...)`, but the production request path builds a plain `Req.new()` from URL/json/headers/timeout options and never injects the `Req.Test` plug/adapter for the `Ollama` test name. Unless the project has hidden global Req test wiring, these tests may still attempt `localhost:11434` instead of the mock.

2. Tool-call execution runs `Enum.map(&execute_tool_call/1)` before reducing results, so all tool calls execute even if an earlier tool call fails. For Lux tools with side effects, that makes failures non-atomic and can run later tools after the first error that should have stopped the response.

3. `execute_tool_call/1` uses `Jason.decode!/1` for model-supplied arguments. A malformed `arguments` string from Ollama would raise and bypass the `{:error, reason}` style used elsewhere in the module. A safe decode path with a fixture for invalid JSON would make tool execution reviewable.

4. `handle_response/2` has no explicit error branch for malformed OpenAI-compatible responses, for example missing `choices`, missing `message`, or missing `finish_reason`. In those cases the `with` expression returns the unmatched value rather than a stable `{:error, ...}` tuple, which can leak inconsistent return shapes to callers.

5. The mocked test path should also prove that application config keys such as `:endpoint`, `:model`, `:temperature`, and `:receive_timeout` do not get passed through to `Req.new/1` as stray request options. Right now provider config and request options are merged late, so a regression test around the final Req options would help.

No live Ollama daemon, network call, API key, or model-generated tool execution should be needed to validate these; mocked Req fixtures and deterministic tool modules should cover the path.

MyTH-zyxeon:
Follow-up2 review-assist pass for #96 / PR #632 at head `60bb059120629dcd5de0f18507d2c6e481d1534d`.

One additional acceptance gap I would close before maintainer review: the PR currently implements an Ollama chat-completions provider, but #96 also asks for local model management, model download/caching, resource controls, and performance monitoring.

Those criteria are not yet represented in the visible diff. `Lux.LLM.Ollama.call/3` posts to `/v1/chat/completions`, but there is no reviewable management surface for listing installed models, pulling/showing a model, recording cache state, enforcing local resource limits, or reporting latency/token/runtime metrics. Because these are side-effect-prone local operations, I would make them explicit and testable with an injectable client/runner: for example, a small `ModelManager` module that can plan/list/pull/show models without requiring a live Ollama daemon in CI, plus fixture tests for cache-hit/cache-miss state, unavailable model errors, timeout/resource-limit errors, and performance metadata in the returned signal.

That would separate "chat provider works" from "local model lifecycle is managed", which is the remaining part of the bounty that the current tests do not yet prove.

No Ollama daemon, model download, local inference call, network model request, API key, or live workflow was used for this static review.

Issue #95, opened 2025-02-25
OpenRouter Integration $500


rparcus:
## Overview

Implement integration with OpenRouter to provide access to a wide range of LLM models through a single unified API.

## Core Features
- OpenRouter API integration
- Multiple model support
- Unified API interface
- Cost optimization
- Error handling and retries

## Technical Details
- Implement OpenRouter client
- Add model selection and routing
- Implement rate limiting and quotas
- Add cost tracking and optimization
- Set up proper error handling

## Acceptance Criteria
- [ ] Implement OpenRouter API client
- [ ] Add support for multiple models
- [ ] Create unified interface matching other providers
- [ ] Implement proper error handling and retries
- [ ] Add cost tracking and reporting
- [ ] Include comprehensive documentation
- [ ] Add integration tests

## Budget: $500

rparcus:
For this, we would like to have a dedicated elixir client. 

- Configuring automatic model selection, fallbacks, provider load balancing, sorting, quantization opts, cache control, etc should be supported via options.
- Streaming is good but not a required goal as the feature is not supported by the main Lux.LLM module yet.

Can I assign this to you @b4s36t4 ?

abhishek818:
@rparcus I can take this up?

abhishek818:
@rparcus I can take this up?

rparcus:
assigned to @b4s36t4. Looking forward for this integration!

Kevinivy00:
After reviewing the detailed requirements, I realize this task requires extensive testing with live APIs and infrastructure setup that goes beyond code writing alone. I'm withdrawing to allow someone with the appropriate dev environment to take this on. Thanks!

daletyler1737:
/opire try

I am submitting PR #472 with a complete OpenRouter provider implementation.

Implemented:
- Lux.LLM.OpenRouter module (follows existing provider pattern)
- Supports 6+ models: claude-3.5-sonnet, gpt-4o, gpt-4o-mini, gemini-pro, llama-3-70b, mistral-7b
- Tool/function calling support
- Automatic retries, streaming, JSON mode
- Full ProviderRegistry integration with per-model cost tracking
- ExUnit tests included

daletyler1737:
Hi @rparcus - I've submitted PR #472 with a complete OpenRouter provider implementation. Happy to address any review feedback. Let me know if you'd like any changes!

zp6:
/claim #95
Working on this issue. Will submit a PR shortly.

[...]

PR #633, opened 2026-05-28
feat: add OpenRouter LLM integration (#95 - multi-provider routing)

JirA44:
## Summary
Implements Lux.LLM.OpenRouter for multi-provider model routing (200+ models).

- OpenAI-compatible API with HTTP-Referer and X-Title headers
- Supports tools, JSON schema, rate limit handling (429)
- Configurable site_url and site_name for attribution
- Unit tests with Req.Test mocks

Closes #95


MyTH-zyxeon:
Review refresh for #95 / PR #633:

This PR is a concrete OpenRouter implementation rather than a design-only note: it adds `Lux.LLM.OpenRouter` plus unit coverage under `lux/test/unit/lux/llm/open_router_test.exs`.

- API client and unified provider interface: `Lux.LLM.OpenRouter` follows the existing LLM-provider shape and keeps the OpenRouter-specific headers/config isolated.
- Multi-model support: model selection is configurable, and the PR exposes model-listing flow instead of hard-coding only one model.
- Error handling/retries: the implementation accounts for auth/rate-limit/error responses, including 429 handling.
- Cost/reporting support: the code includes cost-estimation/model-pricing helpers without requiring a live API key in tests.
- Documentation/tests: the PR body includes usage examples, and the tests use mocks rather than calling OpenRouter with real credentials.

Remaining maintainer-side checks I would treat as merge gates:
- Run the Lux Elixir test target for the new module and confirm the mock setup works in CI.
- Confirm the config key names match the project convention for other `Lux.LLM.*` providers.
- Confirm no live OpenRouter API key, external billing call, or network dependency is required for the test suite.

I did not run any live OpenRouter request or use any API secret for this review pass.

MyTH-zyxeon:
Follow-up review for #95 / PR #633 after re-checking the current diff and acceptance criteria.

The implementation is still a useful concrete OpenRouter provider, but I found a few acceptance-risk items worth tightening before merge:

1. `Config.endpoint` is defined but `call/3` always sends requests to the module `@endpoint`. That makes custom endpoints, test endpoints, proxy routing, or future OpenRouter-compatible hosts impossible even though the config struct advertises endpoint override support.

2. A missing OpenRouter API key is not rejected before the request. `build_headers/1` will still build an Authorization header from `Lux.Config.resolve(api_key)`, so a missing/nil key can become a confusing remote 401 or malformed bearer value instead of a local `{:error, :missing_api_key}` style provider failure.

3. `maybe_add_response_format/2` appends `"Reply in json format"` to every message with `&1.content <> ...`. That assumes every configured message has string content, and it mutates system/assistant/tool messages too. Tool-call and multi-message conversations can crash or have prior messages rewritten before the request.

4. Tool execution can fail noisily or reorder results: `Jason.decode!(args)` raises on malformed tool-call arguments instead of returning `{:error, ...}`, and the reduce path prepends each result, reversing multiple tool call results. Both are painful for a provider expected to support function/tool workflows.

5. The implementation handles auth and 429 responses, but I do not see retry/backoff, quota/rate-limit tracking, or cost reporting wired into the returned metadata. Those are explicit #95 acceptance criteria, so either implementing them or narrowing the PR scope in the issue would reduce review ambiguity.

No live OpenRouter request, API key, billing flow, or secret-bearing workflow was used for this review.

MyTH-zyxeon:
Follow-up review assist for #95 / PR #633 at head `c5eb92251c5170d53b4c6b2a7be64dfd5bc21204`.

One additional CI-facing issue I would fix before merge:

- The new `OpenRouterTest` uses `Req.Test.expect(OpenRouter, ...)`, but this PR does not update `test/test_helper.exs` / `UnitAPICase` to route `Lux.LLM.OpenRouter` through `plug: {Req.Test, OpenRouter}`. Existing provider tests work because `UnitAPICase` already sets `Application.put_env(:lux, OpenAI, plug: {Req.Test, OpenAI})`, `Anthropic`, `TogetherAI`, etc., and each provider merges `Application.get_env(:lux, __MODULE__, [])` into the Req options. Without the matching `Application.put_env(:lux, OpenRouter, plug: {Req.Test, OpenRouter})` setup, these tests can miss the mock and attempt the real OpenRouter endpoint or fail before exercising the expected conn assertions.

This is separate from the earlier endpoint/API-key/json/tool-call/cost-tracking notes. I did not run a live OpenRouter request, use any API key, or execute billing/network-dependent code.

Issue #99, opened 2025-02-25
LLM Provider Abstraction Layer $600


rparcus:
## Overview

Implement a universal provider interface to manage multiple LLM providers efficiently, with automatic model selection, fallback handling, and optimization features.

## Core Features
- Universal provider interface
- Automatic model selection
- Smart fallback handling
- Cost and performance monitoring
- Caching and optimization

## Technical Details
- Design and implement universal interface
- Create provider registry system
- Implement selection logic
- Add monitoring capabilities
- Set up optimization features

## Acceptance Criteria
- [ ] Design and implement universal provider interface
- [ ] Create provider registry and management system
- [ ] Implement automatic model selection logic
- [ ] Add intelligent fallback handling
- [ ] Implement cost tracking and optimization
- [ ] Add performance monitoring and analytics
- [ ] Include documentation and examples

## Budget: $600

sushi-levanter:
Hi! Built the LLM provider abstraction layer in PR #376 — unified interface for OpenAI, Anthropic, Google, and local models with streaming, function calling, and cost tracking. 8 modules, 92 tests.

Happy to iterate based on your review!

solosofi:
## Claiming: LLM Provider Abstraction Layer ($600)

I've implemented the full LLM provider abstraction system.

**PR:** #?

### What's Implemented

**LlmProviderAbstractionPrism** — Universal LLM interface supporting:
- 5 providers: OpenAI, Anthropic, Google AI, Ollama, OpenRouter
- Auto model selection with preference-based routing
- Smart fallback chain on provider failure
- Cost estimation and enforcement per request
- Token usage and latency monitoring

Follows existing Lux.Prism + Lux.Python + ~PY pattern.

**Wallet for payment (Base):** `0x2d44fc27a616606b42448309F4d8e3F423d93267`

solosofi:
## Claiming: LLM Provider Abstraction Layer ($600)

I've implemented the full LLM provider abstraction system.

**PR:** #475 (includes both NFT and LLM work)

### What's Implemented

**LlmProviderAbstractionPrism** — Universal LLM interface supporting:
- 5 providers: OpenAI, Anthropic, Google AI, Ollama, OpenRouter
- Auto model selection with preference-based routing
- Smart fallback chain on provider failure
- Cost estimation and enforcement per request
- Token usage and latency monitoring

Follows existing Lux.Prism + Lux.Python + ~PY pattern.

[...]

PR #634, opened 2026-05-28
feat: LLM Provider abstraction — routing, fallback, cache, cost tracking (#99)

JirA44:
## Summary
Implements Lux.LLM.Provider — a universal router over all registered LLM backends.

**Features:**
- Provider registry with priority, cost_per_1k, and capability tags
- Selection strategies: :priority | :cost | :latency | :random
- Automatic fallback chain when a provider fails
- ETS-backed response cache with configurable TTL (0 to disable)
- Per-provider stats: calls, errors, avg_latency_ms, total_cost
- Force a specific provider: %{provider: Lux.LLM.Ollama}

**Usage:**
`lixir
# Auto-select cheapest cloud provider with tool support
Lux.LLM.Provider.call("Summarise PR", tools, %{tags: [:cloud, :tools], selection: :cost})

# Check costs
Lux.LLM.Provider.stats()
# => %{Lux.LLM.OpenAI => %{calls: 12, total_cost: 0.24, avg_latency_ms: 820}}
`

Closes #99


MyTH-zyxeon:
Nice direction for #99. I think there is one acceptance-risk worth tightening before review:

`load_config/1` reads `selection`, `fallback`, `cache_ttl`, and `track_costs` only from application defaults, while the call examples and tests pass per-call config maps. That means calls like:

```elixir
Lux.LLM.Provider.call("Summarise PR", tools, %{tags: [:cloud], selection: :cost, cache_ttl: 300})
```

will keep the app-default selection/cache/fallback settings instead of the caller's requested strategy. Only `tags` is currently taken from overrides. For this bounty's "automatic model selection", "smart fallback", and "caching/optimization" acceptance criteria, it would be stronger to merge those override keys into `cfg` and add tests that prove:

- `selection: :cost` changes provider order,
- `fallback: false` stops after the first provider error,
- `cache_ttl` actually reuses a successful provider response.

That would turn the current structural tests into behavior-level coverage for the main router contract.

MyTH-zyxeon:
Thanks, I rechecked PR #634 at head `df9dfaefc788bcb6a6a3d19a08226e99315becf9`. A few more implementation-level issues stood out beyond my earlier config-override note:

- The cache key is only `phash2({prompt, provider.module})`. It ignores `tools`, model/options, tags, and caller config, so two calls with the same prompt but different tools or model constraints can receive the wrong cached response.
- `filter_by_tags/2` uses `Enum.any?/2`, so a request like `tags: [:cloud, :tools]` accepts providers that have only one of those tags. For capability routing this should probably require all requested tags, or the API should document OR semantics explicitly.
- `record_stats/4` looks up cost from `@default_providers`, not the active configured provider list. Any application-level provider registry or `cost_per_1k` override will produce incorrect cost totals.
- When `fallback: false`, provider errors return through the `{:error, _} = err` branch without `record_stats/4`, so monitoring undercounts failures for the exact mode that is meant to stop on the first failure.
- Forced provider calls (`%{provider: Mod}`) bypass cache, fallback, stats, and cost tracking entirely. If that is intentional, it needs to be documented; otherwise the forced-provider path should still go through the same instrumentation boundary.
- Cache and stats share one public ETS table. Tests only call `clear_cache/0`, which deletes cache entries but leaves stats behind; repeated tests or long-lived nodes can carry monitoring state across scenarios without a `clear_stats/0` or reset helper.

Suggested next slice: include `tools` and routed model/options in the cache key, tighten tag matching semantics, derive cost from the active provider definition, record failed non-fallback attempts, and add behavior tests for cache isolation, configured cost overrides, forced-provider instrumentation, and stats reset.

MyTH-zyxeon:
Rechecked #634 again at head `df9dfaefc788bcb6a6a3d19a08226e99315becf9`. One more acceptance-risk area I would tighten before this is review-ready: the router still needs an integration-level contract against the existing `Lux.LLM` providers, not only the new wrapper module.

Specific issues I see:

- `@default_providers` includes `Lux.LLM.OpenRouter` and `Lux.LLM.Ollama`, but the current `main` tree has `Lux.LLM.OpenAI`, `Lux.LLM.Anthropic`, `Lux.LLM.TogetherAI`, and `Lux.LLM.Mira`. If those modules are not added elsewhere, a routed fallback path can hit an undefined provider module at runtime instead of producing a normal `{:error, reason}` and continuing.
- Routed calls drop caller options before reaching the selected provider. `provider_config/2` only merges `provider.config` with `%{track_costs: ...}`, so normal router calls do not pass through existing `Lux.LLM.OpenAI` options such as `model`, `temperature`, `messages`, `json_schema`, `tool_choice`, or timeouts unless every value is duplicated inside the provider descriptor.
- The docs show model selection as a router feature, but the default provider descriptors have no model/capability metadata and the selected provider config does not inject a model. At the moment this looks like provider-module selection rather than automatic model selection.
- `try_providers/5` assumes provider failures are returned as `{:error, reason}`. Existing provider modules can also raise during request preparation, for example unsupported tool conversion in `Lux.LLM.OpenAI.tool_to_function/1`. A raised exception would bypass the smart fallback path unless the router normalizes provider calls at that boundary.
- The tests do not exercise the real behaviour contract: no fake provider proves that routed calls receive the original options, no fallback test proves an exception or bad provider module advances to the next candidate, and no test verifies that the router returns `{:ok, %Lux.LLM.Response{}} | {:error, _}` consistently.

Suggested next slice: replace or feature-gate missing default modules, pass caller options through the routed provider config, add explicit model/capability metadata if model selection is in scope, wrap provider invocation so fallback handles normalized failures, and add behaviour-level tests with two stub providers where the first fails and the second receives the exact prompt/tools/options.

Issue #58, opened 2025-02-25
Twitter API Core Integration ($2,500)


rparcus:
# Twitter API Core Integration

## Overview
Implement core Twitter API integration for Lux, enabling authentication, tweet management, and user interactions.

## Core Features
- OAuth 2.0 authentication flow
- Tweet creation, deletion, and editing
- Quote tweets and thread creation
- Media upload support (images, videos, GIFs)
- User profile management
- Rate limit handling

## Acceptance Criteria
- [ ] Complete Twitter API v2 integration
- [ ] OAuth flow implementation with token management
- [ ] Comprehensive error handling
- [ ] Rate limit management
- [ ] Documentation and examples
- [ ] Test coverage > 90%
- [ ] Performance benchmarks

## Budget: $2,500

DicksonWu654:
I can do this

rparcus:
That would be awesome! Once we merge some other work we can assign this to you. Until then, it remains open to others who want to engage and take on the work.

Joegold584:
I'm happy to do this!

jaykayudo:
Hello @rparcus, i will like to take this on!

Joegold584:
Hey Jay. I’ve already started working on it and will be releasing a PR
soon. Sorry about that. Maybe you can work on the Discord issues instead?

mudaly10-coder:
Auto-claim: taking this on (HAF).
Will submit a minimal PR quickly; happy to align scope/acceptance if needed.

addidea:
I'd like to work on this bounty. I have experience with Twitter API v2 and OAuth 2.0 implementations. My fork: https://github.com/addidea/lux

Suraj-kumar00:
Hi @rparcus,

I’d love to work on this issue. It looks quite interesting, and I’m excited to implement it.

From your earlier comment, I understand that it won’t be assigned immediately and will remain open for contributors. Just to confirm, is that correct?

I’ve started looking into it and plan to raise a PR soon if you could confirm that this issue is still active and that contributions are welcome? I noticed there are currently no labels on it other then bounty, so I just want to make sure it’s still in scope.

Also, I haven’t seen much activity on the project over the past few months. Just checking, is the project still actively maintained?

Thanks!

Fred-Zhang83:
我可以来做这个（如果确实是有偿/悬赏任务的话）。为了对齐预期并尽快开始，麻烦确认几件事：

[...]

PR #635, opened 2026-05-28
Add Twitter API v2 core integration

sureshchouksey8:
## Summary

Implements the core X/Twitter API v2 integration requested in #58 with a narrow Lux-native package:

- OAuth 2.0 PKCE authorization URL and token/refresh helpers
- agent-facing OAuth prisms for authorization URL building and token exchange
- tweet create, delete, edit, quote, thread, lookup, and recent search helpers
- chunked media upload steps: INIT, APPEND, FINALIZE, STATUS
- user lookup, authenticated-profile lookup, followers/following reads, follow, and unfollow helpers
- structured rate-limit handling: parsed limit/reset/retry-after metadata plus explicit `{:rate_limited, details}` errors for 429 responses
- agent-facing Lux prisms for tweet/media/auth/user actions
- agent-facing Twitter lens modules for tweet/user/social-graph/search reads
- Livebook guide and a small benchmark entry
- mocked `Req.Test` unit coverage with no live Twitter credentials

## Latest hardening

Commit `05512aa` closes two acceptance-risk gaps I found after comparing the bounty criteria and competing PRs:

- upgrades rate limits from raw header exposure to parsed management metadata and structured 429 errors agents can reschedule from
- adds first-class followers/following social-graph reads through both the client and Lux lenses, with mocked tests and guide examples

## Verification

- `mix format --check-formatted`
- `git diff --check`
- `MIX_ENV=prod mix compile` passes locally under conda Elixir 1.19/OTP 29; remaining warnings are pre-existing project warnings outside the Twitter integration
- Targeted HTTP smoke checks against `Req.Test` passed before this hardening for tweet creation, thread chaining, search lens routing, OAuth URL/token exchange, authenticated-profile lookup, and unfollow routing
- Added mocked tests for structured rate-limit errors plus followers/following reads; local `MIX_ENV=test mix test ...` is still blocked before project tests execute by the existing Credo/Elixir mismatch described below

Full `mix test` could not be completed in this local environment because the available conda runtime is Elixir 1.19/OTP 29 while the repository pins Elixir 1.18/OTP 27; the test/dev dependency `credo` fails during dependency compilation before the new tests run. The implementation and tests are written to the repository's existing `UnitAPICase`/`Req.Test` style so CI on the pinned runtime should exercise them normally.

Refs #58

### DIFF
--- a/lux/benchmarks/twitter_client_benchmark.exs
+++ b/lux/benchmarks/twitter_client_benchmark.exs
@@ -1,0 +1,17 @@
+Mix.install([{:benchee, "~> 1.3"}])
+
+alias Lux.Integrations.Twitter.Client
+
+Benchee.run(%{
+  "authorization_url/1" => fn ->
+    Client.authorization_url(%{
+      client_id: "client",
+      redirect_uri: "https://example.com/callback",
+      state: "state",
+      code_verifier: "verifier"
+    })
+  end,
+  "pkce_pair/1" => fn ->
+    Client.pkce_pair("dBjftJeZ4CVP-mB92K27uhbUJU1p1r_wW1gFWFOEjXk")
+  end
+})
--- a/lux/config/runtime.exs
+++ b/lux/config/runtime.exs
@@ -22,6 +22,7 @@ if config_env() in [:dev, :test] do
     openweather: env!("OPENWEATHER_API_KEY", :string!, "missing open weather"),
     transpose: env!("TRANSPOSE_API_KEY", :string!, "missing transpose"),
     discord: env!("DISCORD_API_KEY", :string!, "missing discord"),
+    twitter_bearer: env!("TWITTER_BEARER_TOKEN", :string!, required: false),
     etherscan: env!("ETHERSCAN_API_KEY", :string!, "missing etherscan"),
     etherscan_pro: env!("ETHERSCAN_API_KEY_PRO", :string!, required: false) == "true",
     telegram_bot: env!("TELEGRAM_BOT_TOKEN", :string!, required: false),
--- a/lux/config/runtime.exs
+++ b/lux/config/runtime.exs
@@ -28,7 +28,8 @@ if config_env() in [:dev, :test] do
     telegram_bot: env!("TELEGRAM_BOT_TOKEN", :string!, required: false),
     integration_openai: env!("INTEGRATION_OPENAI_API_KEY", :string!, "missing open ai"),
     integration_anthropic: env!("INTEGRATION_ANTHROPIC_API_KEY", :string!, "missing anthropic"),
-    integration_openweather: env!("INTEGRATION_OPENWEATHER_API_KEY", :string!, "missing open weather"),
+    integration_openweather:
+      env!("INTEGRATION_OPENWEATHER_API_KEY", :string!, "missing open weather"),
     integration_transpose: env!("INTEGRATION_TRANSPOSE_API_KEY", :string!, "missing transpose"),
     integration_discord: env!("INTEGRATION_DISCORD_API_KEY", :string!, "missing discord"),
     integration_telegram_bot: env!("INTEGRATION_TELEGRAM_BOT_TOKEN", :string!, required: false),
@@ -37,6 +38,10 @@ if config_env() in [:dev, :test] do
   config :lux, Lux.Integrations.Allora,
     base_url: env!("ALLORA_BASE_URL", :string!, "https://api.upshot.xyz/v2"),
     chain_slug: env!("ALLORA_CHAIN_SLUG", :string!, "testnet")
+
+  config :lux, Lux.Integrations.Twitter,
+    api_url: env!("TWITTER_API_URL", :string!, "https://api.x.com"),
+    authorize_url: env!("TWITTER_AUTHORIZE_URL", :string!, "https://x.com/i/oauth2/authorize")
 
   config :lux, :accounts,
     wallet_address: env!("WALLET_ADDRESS", :string!),
@@ -61,12 +66,14 @@ end
 if config_env() == :test do
   # Add Hammer configuration
   config :hammer,
-    backend: {Hammer.Backend.ETS,
-      [
-        expiry_ms: 60_000 * 60 * 4,       # 4 hours
-        cleanup_interval_ms: 60_000 * 10,  # 10 minutes
-        pool_size: 1,
-        pool_max_overflow: 2
-      ]
-    }
+    backend:
+      {Hammer.Backend.ETS,
+       [
+         # 4 hours
+         expiry_ms: 60_000 * 60 * 4,
+         # 10 minutes
+         cleanup_interval_ms: 60_000 * 10,
+         pool_size: 1,
+         pool_max_overflow: 2
+       ]}
 end
--- a/lux/guides/twitter_api_core.livemd
+++ b/lux/guides/twitter_api_core.livemd
@@ -3,6 +3,27 @@
 This guide shows how Lux agents can use X/Twitter API v2 without putting live
 credentials in tests or examples.
 
+## Runtime Configuration
+
+The client defaults to the current X API host, `https://api.x.com`. If an app
+or account still needs the legacy Twitter host for v2 endpoints, set
+`TWITTER_API_URL=https://api.twitter.com` or configure the integration directly:
+
+```elixir
+config :lux, Lux.Integrations.Twitter,
+  api_url: "https://api.x.com",
+  authorize_url: "https://x.com/i/oauth2/authorize"
+```
+
+Minimal credential shapes by mode:
+
+- App-only reads: `TWITTER_BEARER_TOKEN`
+- OAuth 2.0 PKCE: `TWITTER_CLIENT_ID`, `TWITTER_REDIRECT_URI`, generated
+  verifier/challenge, short-lived `access_token`, and optionally
+  `TWITTER_CLIENT_SECRET` plus `refresh_token`
+- Write/media/follow operations: a user OAuth `access_token`; app-only bearer
+  tokens are intentionally rejected for mutation paths
+
 ## OAuth 2.0 PKCE
 
 ```elixir
@@ -117,6 +138,42 @@ Lux.Integrations.Twitter.Client.search_recent(
   %{access_token: tokens["access_token"], with_rate_limit: true}
 )
 ```
+
+## Media Upload
+
+Chunked media upload is exposed one step at a time so agents can persist the
+server response between calls:
+
+```elixir
+{:ok, init} =
+  Lux.Prisms.Twitter.Media.UploadMedia.handler(
+    %{
+      action: :init,
+      total_bytes: 1_024_000,
+      media_type: "video/mp4",
+      media_category: "tweet_video",
+      access_token: tokens["access_token"]
+    },
+    %{}
+  )
+
+media_id = get_in(init, ["data", "id"]) || init["media_id_string"] || init["media_id"]
+
+Lux.Prisms.Twitter.Media.UploadMedia.handler(
+  %{action: :append, media_id: media_id, segment_index: 0, media: chunk, access_token: tokens["access_token"]},
+  %{}
+)
+
+Lux.Prisms.Twitter.Media.UploadMedia.handler(
+  %{action: :finalize, media_id: media_id, access_token: tokens["access_token"]},
+  %{}
+)
+```
+
+For status polling, pass only the returned `media_id`. Response shapes can
+include either nested `data.id` values or legacy-style `media_id_string`
+depending on the endpoint/account behavior, so production agents should store
+the full response payload as well as the extracted media ID.
 
 ## Tests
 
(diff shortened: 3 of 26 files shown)


MyTH-zyxeon:
Nice broad slice for #58, especially keeping the tests on `Req.Test` and avoiding live Twitter credentials.

One acceptance-risk I noticed around OAuth token exchange: the test asserts that the request body decodes as a form, but it does not assert the outbound `content-type`. `headers/1` defaults every request to `Content-Type: application/json`, while `token_opts/2` only adds `:form` and optional Basic auth. If Req does not override that explicit header for `:form`, X's `/2/oauth2/token` endpoint can reject the request even though the mocked body is correct.

Could you add a small assertion in the token exchange test that `Plug.Conn.get_req_header(conn, "content-type")` is `application/x-www-form-urlencoded` (or adjust `headers/1` so form/multipart requests don't inherit the JSON content type)? The media upload test already checks multipart content type, so adding the same guard for OAuth would make the credential-free coverage stronger.

sureshchouksey8:
Thanks for catching that. Fixed in `b882173`: form and multipart requests no longer inherit the JSON content type, and the OAuth token-exchange test now asserts `application/x-www-form-urlencoded` explicitly. Re-ran `mix format --check-formatted`, `git diff --check`, and `MIX_ENV=prod mix compile` successfully; the local `MIX_ENV=test mix test ...` command is still blocked before tests execute by the repo's existing Credo/Elixir 1.19 mismatch.

MyTH-zyxeon:
Follow-up after rechecking PR #635 for bounty #58. Thanks for addressing the earlier OAuth token content-type issue; I am not claiming the bounty here, just leaving another review-assist pass on the current diff.

- The default OAuth scopes include tweet/user/follow permissions and `offline.access`, but not `media.write`. The PR advertises chunked media upload support, and current X API docs map `media.write` to media upload permission. Without that scope in the default authorization flow, a user can complete OAuth and still fail the media-upload path.
- The media upload client still appears to model a command-based flow against `/2/media/upload` with `command: INIT | APPEND | FINALIZE | STATUS`. Current X API v2 docs expose separate endpoints such as `/2/media/upload/initialize`, `/2/media/upload/{id}/append`, `/2/media/upload/{id}/finalize`, and status lookup by `media_id`. The tests currently assert the command-style request body, so they can pass while the endpoint contract remains incompatible with the current v2 docs.
- `media_upload(:status, ...)` sends both `command=STATUS` and `media_id` to `/2/media/upload`; the current status docs only require `media_id`. I would add a test that matches the documented status query exactly.
- `media_upload(:append, ...)` builds multipart fields with `media_id` in the body rather than the upload id in the path. If the implementation is intentionally targeting an older/alternate protocol, the guide should state that; otherwise the append/finalize path likely needs to follow the documented v2 media-upload URL shapes.
- The OAuth/token pieces now look better after the form fix, but there is still no test proving an authorization URL with the default scopes can cover every feature exposed by the prisms, especially media upload.
- The structured rate-limit handling is useful; I would also cover 403 permission/scope errors for tweet write and media upload so agents can distinguish missing scope from retryable throttling.

A focused next pass would update the default scopes, align media upload paths and tests with the current X API v2 media docs, and add missing-scope/permission tests for media and write actions.

MyTH-zyxeon:
Follow-up review assist on the current #635 diff after the OAuth content-type fix and the later media-upload review. I am still not claiming #58; this is another acceptance-risk pass for the maintainer/author.

- The mutation prisms (`CreateTweet`, `DeleteTweet`, `EditTweet`, `FollowUser`, `UnfollowUser`, media upload) do not require `access_token`, and `Client.headers/1` falls back to the configured `twitter_bearer` token. That can let write/user-interaction paths attempt app-only bearer auth instead of a user OAuth token. Please require an explicit user access token for write/follow/media prisms and add a test that missing access_token fails before any request is sent.
- `ExchangeToken` only requires `grant_type` and `client_id` in the schema. The authorization-code path can therefore send a token request without `code`, `redirect_uri`, or `code_verifier`, and refresh can send without `refresh_token` because nil fields are dropped. Please preflight those grant-specific required fields and return a stable error tuple.
- `normalize_grant_type/1` has no fallback for an unsupported string. If runtime schema validation is bypassed or incomplete, `"client_credentials"`/unknown values can raise a function-clause error instead of returning `{:error, ...}`.
- Thread creation posts sequentially and returns only the failing reason if a later post fails. Since the first tweet may already be live, the error contract should include partial responses or the last published id so an agent can retry or clean up safely.
- The rate-limit metadata wrapper is only applied on successful responses and the dedicated 429 branch. Permission/scope errors such as 403 still lose rate-limit headers and structured retry/diagnostic context. A non-live Req.Test case for 403 would make the write-path error handling clearer.

These are all credential-free checks, but they would make the bounty acceptance story much stronger around OAuth token management, user interactions, and comprehensive error handling.

sureshchouksey8:
Thanks for the second pass. I pushed `b2c6da6` to close both acceptance-risk points:

- default OAuth scopes now include `media.write`, so the generated authorization URL covers the chunked media upload path advertised by the PR
- tweet/media/follow mutation paths now require an explicit user `access_token` and return `{:error, :access_token_required}` instead of silently falling back to the app bearer token for write/user-interaction requests

Validation rerun:
- `mix format --check-formatted`
- `git diff --check`
- `MIX_ENV=prod mix compile`

Targeted `MIX_ENV=test mix test test/unit/lux/integrations/twitter/client_test.exs test/unit/lux/prisms/twitter/tweets_test.exs` is still blocked before project tests execute by the repo's existing Credo/Elixir 1.19 regex compatibility issue, but I added focused assertions for `media.write` scope and app-only bearer rejection on write/media/follow operations.

sureshchouksey8:
Follow-up hardening pushed in `842ac56` for the remaining credential-free acceptance risks from the review-assist pass.

What changed:
- aligned chunked X API v2 media upload routing to `POST /2/media/upload/initialize`, `POST /2/media/upload/{id}/append`, `POST /2/media/upload/{id}/finalize`, and status lookup by `media_id`
- added media-id/media preflight errors for append/finalize/status instead of building invalid paths
- added grant-specific OAuth token preflight errors for missing authorization-code and refresh-token fields
- added unsupported grant-type handling instead of function-clause crashes
- preserved partial thread progress when a later tweet in a thread fails
- preserved optional rate-limit metadata on 403/scope-style API errors when `with_rate_limit` is requested

Validation run locally:
- `mix format lib/lux/integrations/twitter/client.ex lib/lux/prisms/twitter/auth/exchange_token.ex test/unit/lux/integrations/twitter/client_test.exs test/unit/lux/prisms/twitter/tweets_test.exs`
- `git diff --check`
- `MIX_ENV=prod mix compile`

Targeted test command is still blocked in this local Elixir 1.19 environment before test execution by the existing Credo dependency regex compile error in `lib/credo/check/config_comment_finder.ex`. The new coverage is committed in the repo's existing `Req.Test` style so pinned CI can exercise it once dependencies compile in the project runtime.

MyTH-zyxeon:
Review-assist (Twitter API v2 core integration):

1) CI / checks
- `mergeStateStatus=BLOCKED` + `reviewDecision=REVIEW_REQUIRED` is expected, but there are **no checks reported** for the branch right now (`gh pr checks` shows none). If the repo expects CI, it may simply not be configured for this branch/PR yet — but it’s worth confirming so reviewers can rely on test signals.

2) Base URL compatibility (`api.x.com` vs `api.twitter.com`)
- I see the client defaults to `https://api.x.com` (`Lux.Integrations.Twitter.api_url/0`). Depending on account/app setup, some deployments still expect `https://api.twitter.com` for v2 endpoints.
- Suggestion: keep the default, but make it explicitly configurable via env (e.g. `TWITTER_API_URL`) or document the `Application.get_env(:lux, Lux.Integrations.Twitter)` override path in `runtime.exs`/the Livebook.

3) Credentials and token shapes
- `TWITTER_BEARER_TOKEN` is optional in `runtime.exs` (good for tests). If production usage is expected, consider also documenting the full minimal env set for each mode:
  - app-only: bearer token
  - OAuth2 PKCE: client_id (+ optional client_secret) + redirect_uri + scopes + short-lived access_token (+ refresh_token)

4) Rate-limit handling
- The structured `{:error, {:rate_limited, %{rate_limit: ...}}}` path + `with_rate_limit: true` metadata looks great.
- Small edge: some responses omit `retry-after`; you already guard via `parse_int/1` — looks safe.

5) Media upload
- Nice to see initialize + upload + finalize paths covered and unit-tested.
- If you run into real-world failures, the most common mismatch is required params (`media_category`, `media_type`, `total_bytes`) and output shape (`media_id_string` vs `media_id`). Might be worth adding a brief note in the Livebook about expected inputs/outputs for each upload step.

Overall: the surface area here is substantial, but the API layout (integrations + lenses + prisms) and Req.Test-driven unit tests are a strong foundation. Happy to help review deeper if you want a focused checklist per operation (auth / tweets / users / search / media).

sureshchouksey8:
Follow-up after getting the local test path unblocked.

I patched the local Credo 1.7.11 dependency regex only under `deps/` to get past its Elixir 1.19 compile issue, then ran the targeted unit suite for this PR.

Pushed `227d6fc` with two small corrections found by that run:
- fixed the Req multipart tuple shape for chunked media append (`{:media, {value, filename: ...}}`)
- corrected the expected UTC timestamp for the existing `1770000000` rate-limit reset header assertions

Validation now passes locally:
- `mix format lib/lux/integrations/twitter/client.ex test/unit/lux/integrations/twitter/client_test.exs`
- `git diff --check`
- `MIX_ENV=test mix test test/unit/lux/integrations/twitter/client_test.exs test/unit/lux/prisms/twitter/tweets_test.exs --include unit`

Result: `26 tests, 0 failures`.

sureshchouksey8:
Follow-up for the review-assist notes around runtime configuration and media docs.

Pushed `7b89141` with:
- explicit `TWITTER_API_URL` / `TWITTER_AUTHORIZE_URL` runtime configuration, defaulting to `https://api.x.com` and `https://x.com/i/oauth2/authorize`
- a unit test proving the client honors a configured `https://api.twitter.com` API base URL
- Livebook documentation for app-only vs OAuth2 PKCE credential shapes, write-token requirements, configurable API host, and chunked media upload input/output handling (`data.id`, `media_id_string`, or `media_id`)

Validation:
- `mix format config/runtime.exs test/unit/lux/integrations/twitter/client_test.exs`
- `git diff --check`
- `MIX_ENV=prod mix compile`
- `MIX_ENV=test mix test test/unit/lux/integrations/twitter/client_test.exs test/unit/lux/prisms/twitter/tweets_test.exs --include unit`

Result: `27 tests, 0 failures`.

MyTH-zyxeon:
Follow-up review-assist for #58 / PR #635 after the latest author hardening commits. No competing PR or claim from me here.

What materially improved since the earlier review passes:

- The OAuth/token content-type issue is fixed, and the targeted unit path is now unblocked locally.
- Mutation paths now require explicit user OAuth access tokens instead of silently falling back to app-only bearer auth.
- `media.write`, configurable `TWITTER_API_URL` / `TWITTER_AUTHORIZE_URL`, and app-only vs OAuth2 PKCE credential shapes are documented.
- The media upload flow now uses documented initialize/append/finalize/status shapes and has Req.Test coverage for multipart append, media-id paths, and status query handling.
- Error handling now covers unsupported grant types, missing grant-specific fields, partial thread failure, 403/scope errors with rate-limit metadata, and 429 rate-limit payloads.

Remaining maintainer-facing acceptance checks I would still run before payout/merge:

1. Confirm repo CI runs the same targeted unit commands reported by the author (`twitter/client_test.exs` + `twitter/tweets_test.exs`) without the local-only Credo dependency patch.
2. Add or verify one lens-level read test that uses string-key schema input, `access_token`, and `with_rate_limit` together so agent/tool invocation does not depend only on atom-key Elixir calls.
3. Decide whether `TWITTER_API_URL=https://api.twitter.com` should be a documented compatibility option only or part of the official acceptance matrix.
4. Keep live tweet posting, follow/unfollow, media upload, OAuth credential use, and real X API requests outside review unless maintainers explicitly opt into a live smoke test account.

At this point the PR looks much closer to acceptance than it did in the earlier passes; the remaining questions are CI/runtime compatibility and final maintainer scope choice, not obvious credential-free implementation gaps.

sureshchouksey8:
Follow-up on the remaining review-risk checklist:

- Pushed `db814ba` to cover schema/tool-style lens input. The Twitter read lenses now normalize known string keys before dispatching to the client, so agent/tool invocation is not dependent on atom-key Elixir calls.
- Added a lens-level regression for `SearchRecent.focus/1` using string-key input with `"query"`, `"access_token"`, `"with_rate_limit"`, and `"plug"`, asserting the bearer token header and returned rate-limit metadata.
- `TWITTER_API_URL=https://api.twitter.com` is intended as a documented compatibility/runtime override. The default remains `https://api.x.com`; I would only add live `api.twitter.com` smoke coverage if maintainers want that in the acceptance matrix.

Current head is `db814ba`; GitHub reports the PR open, non-draft, and mergeable.

Issue #94, opened 2025-02-25
Rust Core Integration Setup $500


rparcus:
## Overview

Implement core Rust language integration for Lux, enabling high-performance native code execution and FFI bindings.

## Core Features
- Basic Rust runtime integration
- Cargo project structure in `priv/rust`
- Type conversion system between Elixir and Rust
- Error handling framework
- Basic FFI bindings

## Technical Details
- Set up Rust project structure following Lux conventions
- Implement NIF bindings for core functionality
- Create type conversion layer for primitive types
- Establish error propagation system
- Add memory safety guarantees

## Acceptance Criteria
- [ ] Initialize Rust project in `priv/rust` with proper structure
- [ ] Implement basic type conversion between Elixir and Rust
- [ ] Create error handling framework with proper propagation
- [ ] Set up FFI bindings with memory safety
- [ ] Add documentation for Rust integration
- [ ] Include example usage in guides
- [ ] Add integration tests

## Budget: $500

cybeleri:
I've submitted PR #352 implementing Rust Core Integration. The implementation includes:

- Rust NIF bindings via Rustler  
- Type-safe conversions between Elixir and Rust
- Error handling framework
- Documentation and integration tests

I would like to formally request assignment to this issue and continue with related Rust bounties (#100, #101, #102) as discussed in the bounty guidelines. Happy to schedule a scope discussion.

Looking forward to your review\!

akash-R-A-J:
Hi @rparcus — I’m interested in working on this issue. Let me know if I can pick it up.

andygoodluck:
Hi @Spectral-Finance team,

I'd like to claim this bounty. I have experience with Rust and Elixir integration using NIFs.

**My approach:**
1. Set up Rust project structure following Lux conventions
2. Implement NIF bindings using rustler
3. Create type conversion for primitives
4. Build error handling framework
5. Add memory safety and tests

**Timeline:** 5-7 days

Ready to start immediately!

Kevinivy00:
Hi! I'm excited to work on this issue. I have strong experience with the relevant technologies and can deliver high-quality work. This would be a great challenge for me. Please assign this to me!

My GitHub: https://github.com/Kevinivy00
Portfolio: 23 repositories showcasing my work

Thanks!

eulicesl:
Hi @rparcus — I’d like to take this one on (if still available).

Execution plan:
[...]

PR #648, opened 2026-05-28
feat: Rust core integration with NIF support ($500 bounty #94)

JirA44:
## Summary

- Adds `Lux.Rust` — inline Rust eval via temporary Cargo projects (`eval/2`, `compile/2`, `with_deps/3`), plus `version/0` and `available?/0` for runtime detection
- Adds `Lux.Rust.NIF` — safe NIF loading via `load/2`, load-state inspection via `loaded?/1`, and crash-isolated NIF calls via `safe_call/3`
- Adds `Lux.Rust.FFI` — pure-Elixir bidirectional type conversion between Elixir and Rust-compatible NIF terms (`to_erlang_term/1`, `from_erlang_term/1`, `encode_for_nif/2`, `decode_nif_result/1`)
- Adds `priv/rust/lux_nif/` — Rustler Cargo project scaffold with `ping`, `add_integers`, `reverse_string`, `abs_list` NIF stubs and a `types.rs` helper module
- Adds `test/unit/lux/rust_test.exs` — comprehensive unit tests covering availability checks, eval round-trips (guarded by `Rust.available?()`), and full FFI type encoding/decoding round-trips

## Test plan

- [ ] `mix test test/unit/lux/rust_test.exs` passes on a machine with `cargo`/`rustc` in PATH
- [ ] `available?/0` and `version/0` tests pass regardless of whether Rust is installed
- [ ] FFI round-trip tests pass without Rust installed (pure-Elixir module)
- [ ] `eval/2` hello-world test produces `"Hello, World!\n"` when `cargo` is available
- [ ] `compile/2` returns `{:ok, binary_path}` for valid code
- [ ] `encode_for_nif/2` correctly rejects out-of-range integers and wrong types
- [ ] `decode_nif_result/1` handles both tagged maps and raw Elixir pass-through values

/opire try


zhaog100:
/opire try

Claiming this bounty. Will deliver a complete implementation.

MyTH-zyxeon:
Review assist for #94 / PR #648:

This is a concrete Rust-core implementation attempt for the #94 bounty: it adds `Lux.Rust`, `Lux.Rust.NIF`, `Lux.Rust.FFI`, a Rustler scaffold under `priv/rust/lux_nif`, and unit tests.

- Rust runtime integration: `Lux.Rust.available?/0`, `version/0`, `eval/2`, `compile/2`, and `with_deps/3` cover the runtime-detection and inline Cargo-project path.
- Cargo project structure: the PR adds a `priv/rust/lux_nif` Cargo project with Rustler dependency and release profile.
- NIF/FFI surface: the PR includes NIF loading/calling wrappers plus pure-Elixir type encoding/decoding helpers.
- Tests: the new test file covers availability, compile/eval guards, FFI round-trips, typed NIF encoding, and NIF-result decoding.

Merge gates I would check before accepting:
- `Lux.Rust.NIF.load/2` accepts a `module` argument but calls `:erlang.load_nif/2` from `Lux.Rust.NIF`; confirm the intended module actually owns the loaded NIF callbacks, or remove/clarify the unused module parameter.
- `Lux.Rust.executable?/1` uses bitwise `&&&`; confirm the project/compiler version imports or qualifies that operator, otherwise this can fail at compile time.
- `Lux.Rust.with_deps/3` discards stderr on `cargo run` and then runs `cargo build` to collect errors; that can lose runtime stderr. Consider capturing combined output from the failing run directly.
- The test plan still has unchecked boxes in the PR body. Maintainer CI should run at least `mix test test/unit/lux/rust_test.exs` with and without Rust/Cargo available, because several tests are guarded on runtime tool availability.

I did not build the Rust project, run any live native code, or execute side-effecting code for this review pass.

MyTH-zyxeon:
Follow-up review assist for #94 / PR #648 after checking the current diff:

The previous review already covered the broad acceptance-criteria mapping and the `load_nif`/stderr/operator gates. A few additional merge blockers look worth checking before this is accepted:

- The Elixir FFI and Rust `LuxTerm::encode_tagged/1` do not appear to use the same tagged shape. Elixir encodes and decodes atom keys/values like `%{type: :integer, value: 42}`, while the Rust helper currently emits string keys/values like `%{"type" => "integer", "value" => 42}`. That means a Rust-produced tagged term will not match `Lux.Rust.FFI.decode_nif_result/1 -> from_erlang_term/1`.
- The Rust `types.rs` helper is not wired into a `Decoder` implementation for the Elixir tagged maps, and `LuxTerm` has no map variant even though the Elixir module documents map conversion. So the "bidirectional type conversion" layer is still mostly pure-Elixir and not yet a real Elixir<->Rust round trip.
- `rustler::init!("Elixir.Lux.Rust.NIF.Native")` exports NIFs for `Lux.Rust.NIF.Native`, but this PR does not add that Elixir module with Rustler stubs/fallback functions (`ping/0`, `add_integers/2`, etc.). As written, there is no obvious Elixir module that owns and exposes those callbacks for `safe_call/3` to invoke.
- `Lux.Rust.NIF.loaded?/1` falls back to `has_nif_exports?/1`, but that helper currently returns `false` for every export, so modules without a populated `:nifs` attribute will never be reported as loaded.
- `Lux.Rust.FFI.from_erlang_term/1` falls back from `String.to_existing_atom/1` to `String.to_atom/1`. If any decoded value can originate from Rust/native/external input, this can grow the BEAM atom table; for a memory-safety-focused FFI layer, it would be safer to avoid creating arbitrary atoms or make that behavior explicit and bounded.

A useful maintainer validation path would be an integration test that builds the Rust crate, loads it into the intended Elixir stub module, calls `ping/0` and `add_integers/2` through `Lux.Rust.NIF.safe_call/3`, and round-trips at least one tagged value through both the Elixir and Rust conversion paths. I did not execute native code or run the PR branch locally for this pass.

MyTH-zyxeon:
Rechecked #94 / PR #648 at head `ad941dd30f8b2545f6dfaefe34bc6315f7855ae0`. The earlier comments covered the NIF owner/stub path, tagged FFI mismatch, and atom-safety concerns. One more acceptance area I would tighten is the runtime safety boundary around temporary Cargo projects and native execution.

Specific gaps:

- `compile/2` creates a temp Cargo project but never removes it. On success it returns only the compiled binary path, not the project directory or a cleanup handle, so callers have no obvious way to clean `src`, `Cargo.toml`, `target`, and downloaded build artifacts. Repeated compile calls can leave unbounded temp trees.
- `with_deps/3` accepts arbitrary dependency tuples and writes them directly into `Cargo.toml` as `#{name} = "#{version}"`. Crate names, version strings, path/git dependencies, or newline/quote characters are not validated or escaped. For agent-provided input this can generate malformed manifests or allow dependency/build-script behavior that is not called out.
- `eval/2` and `with_deps/3` execute arbitrary Rust through `cargo run --release` under the current user environment. That may be expected for a language bridge, but the docs/tests should make the trust boundary explicit and ideally offer a no-network/no-default-env mode or mark this API as trusted-code-only.
- The NIF crate sets `[profile.release] panic = "abort"`. For a BEAM NIF, an aborting native panic can take down the VM, while `Lux.Rust.NIF.safe_call/3` can only catch Elixir throws/exits/exceptions around `apply/3`; it cannot catch a native segfault or process abort. That is a major distinction for the issue's memory-safety acceptance criterion.
- There is no regression test or documented validation path for cleanup, rejected unsafe dependency metadata, or native panic behavior. Even if full sandboxing is out of scope, tests/docs should prove the intended safety contract rather than implying the wrapper isolates every NIF failure.

Suggested next slice: return a managed compile artifact with an explicit cleanup API (or clean temp dirs on failure), validate/escape Cargo manifest inputs, document trusted-code execution limits, reconsider `panic = "abort"` for the NIF crate, and add tests for temp cleanup plus invalid dependency metadata.

Issue #82, opened 2025-02-25
Hyperliquid Integration and Perpetual Trading $900


rparcus:
## Overview

Implement Hyperliquid integration for perpetual trading with focus on position management and risk control.

## Core Features
- Perpetual trading interface
- Position management
- Leverage control
- Risk monitoring
- Order execution
- Liquidation protection
- PnL tracking
- Margin management

## Acceptance Criteria
- [ ] Order placement and management
- [ ] Position tracking system
- [ ] Risk management implementation
- [ ] Liquidation monitoring
- [ ] Margin management system
- [ ] Documentation and examples
- [ ] Integration tests for trading operations
- [ ] Risk control verification

## Budget: \$900

yw13931835525-cyber:
Claiming this bounty. I can deliver a working skeleton within 48 hours (core modules, integration scaffolding, basic tests, and runbook), then iterate to full delivery with daily updates.

lmee:
I reviewed "Hyperliquid Integration and Perpetual Trading $900" and read all comments before proposing a patch path.

Problem I see: Implement Hyperliquid integration for perpetual trading with focus on position management and risk control.
Root-cause hypothesis: a state flow mismatch between action handling and final UI state.
First patch plan: reproduce the issue against acceptance criteria, isolate the root cause in the relevant module, and submit a focused fix with regression tests.
If this direction matches what you want, I can publish repro notes first so review starts from concrete evidence.

templarx:
I'd like to tackle the Perpetual Trading integration. Experience with perp DEXes (Hyperliquid, GMX, dYdX) and building trading infrastructure. Can implement order types, position management, PnL tracking, and liquidation monitoring. -- @templarx (via Plutus Agent)

Asobu01:
I can take Spectral-Finance/lux#82.
For "Hyperliquid Integration and Perpetual Trading $900", I will start by reproducing the UI/layout issue and isolating the component/state that causes it.
Target: draft PR in ~8h.
I can post a reproducible checklist/log snippet in ~30 minutes so you can verify execution speed before the full PR.
Deliverables: minimal UI fix aligned with current UX; quick before/after validation notes; PR with clear change summary.
@Asobu01

Asobu01:
Quick follow-up on `Spectral-Finance/lux#82`. I am still ready to start immediately and can post the first repro/checklist note before the draft PR if helpful. If you want me on this one, I can begin now. (@Asobu01)

[...]

PR #671, opened 2026-06-02
feat: Hyperliquid integration with order/position/risk/margin management

theherrovn-sys:
## Summary

Implements comprehensive Hyperliquid perpetual trading integration for Spectral-Finance/lux (#82).

#### Main Integration (`lux/lib/lux/integrations/hyperliquid/`)
- **client.ex** — HTTP client for Hyperliquid REST API (info + exchange endpoints)
- **websocket.ex** — WebSocket client for real-time data (allMids, l2Book, trades, user streams)
- **models.ex** — Elixir structs for orders, positions, account state, market data, order book, trades, funding
- **signing.ex** — EIP-712 compatible signing using ExSecp256k1 with Lux.Native fallback
- **order_manager.ex** — GenServer for order lifecycle management (place/cancel/track orders)
- **position_manager.ex** — GenServer for position tracking with PnL calculations
- **risk_manager.ex** — GenServer for liquidation monitoring, stop-loss/take-profit, risk assessment
- **margin_manager.ex** — GenServer for margin operations (add/remove margin, cross/isolated mode, leverage)

#### Tests (`lux/test/unit/lux/integrations/hyperliquid/`)
- 60 unit tests covering all modules with proper mocking

### Technical Details
- Uses `Req` for HTTP (matching existing Discord/Telegram patterns)
- Uses `WebSockex` for WebSocket (available as transitive dependency)
- Uses `ExSecp256k1` for ECDSA signing (already a dependency)
- Uses `Lux.Native` for keccak256 hashing with `:crypto` fallback
- Follows existing patterns from Telegram/Discord integrations
- GenServer-based managers with ETS for high-performance state storage

### Risk Management Features
- Configurable leverage limits and position size constraints
- Portfolio concentration monitoring
- Margin ratio warnings/critical alerts
- Liquidation price proximity detection
- Automatic stop-loss and take-profit order placement

Closes #82


MyTH-zyxeon:
I took a credential-free pass over this PR against the #82 acceptance criteria. I did not use Hyperliquid credentials, open a WebSocket, submit/cancel/modify any order, change leverage or margin, or make live/private API calls.

The structure is much closer to a real implementation than the earlier markdown-only attempts, but I think a few acceptance risks are worth tightening before this is safe to merge:

1. Private exchange actions need an explicit no-live boundary. `Client.query_private/2` signs and posts to `/exchange` whenever `hyperliquid_account_key` is present, and the managers call `Client.place_order/2`, `Client.cancel_order/3`, and `Client.modify_margin/3` without a required dry-run/testnet/paper-mode gate. For a perpetual trading integration, I would add an explicit mode flag and tests proving unit/integration paths cannot place live orders by default.

2. Margin removal and margin-mode changes look mapped to the wrong action. `remove_margin/3` passes `%{is_deposit: false, amount: ...}`, but `Client.modify_margin/3` sends `type: "updateLeverage"` with `leverage: 0` for the non-deposit path. `set_margin_mode/3` also passes mode/is_cross without leverage, so it appears to become another leverage-zero action rather than a safe cross/isolated mode change. I would add payload-level tests for add isolated margin, remove isolated margin, set cross/isolated mode, and update leverage.

3. Order payload construction needs Hyperliquid metadata mapping fixtures before signing. `build_order/1` puts the string coin directly into `a: params[:asset]`; Hyperliquid L1 order actions normally need the exchange asset id plus price/size precision handling. Please add tests that fetch or inject metadata and prove asset id, tick size, lot size, price rounding, and size rounding before signing/submission.

4. The WebSocket/user-stream paths need malformed-event and reconnect coverage. `handle_ws_event(%{"channel" => "user"} = event, state)` reads `event["data"]["fills"]`, which can crash on nil/malformed `data`, and `dispatch/2` calls the handler directly. A bad stream event or user callback should not take down order/position/risk tracking without a clear supervisor/reconnect story.

5. The default ETS tables are public named tables (`:hyperliquid_orders`, `:hyperliquid_positions`). That can collide across repeated manager starts and can leak test/runtime state between processes. Private unnamed tables or caller-supplied unique names would make this safer for tests and multi-account use.

6. The issue asks for docs/examples, integration tests for trading operations, and risk-control verification. The PR appears to add unit tests, but I did not see docs/examples or an integration/risk-control test layer that proves the no-live-order invariant, liquidation monitoring behavior, margin/leverage payloads, and risk-triggered close/reduce-only behavior.

If those are covered, maintainers would have a much stronger path to verify this without risking accidental live trading behavior.

PR #679, opened 2026-06-02
feat: add Binance exchange integration for spot trading

JamesJi79:
Full Binance REST API client with public and signed endpoints.

Closes #84

### DIFF
--- a/lux/lib/lux/integrations/binance.ex
+++ b/lux/lib/lux/integrations/binance.ex
@@ -1,0 +1,56 @@
+
+defmodule Lux.Integrations.Binance do
+  @moduledoc """
+  Binance exchange integration for spot trading and market data.
+  """
+  @base_url "https://api.binance.com"
+
+  def ticker(symbol), do: get("/api/v3/ticker/24hr", %{symbol: String.upcase(symbol)})
+  def order_book(symbol, limit \\ 100), do: get("/api/v3/depth", %{symbol: String.upcase(symbol), limit: limit})
+  def recent_trades(symbol, limit \\ 100), do: get("/api/v3/trades", %{symbol: String.upcase(symbol), limit: limit})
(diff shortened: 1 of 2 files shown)


MyTH-zyxeon:
Maintainer-facing review assist for #84 / PR #679. No competing claim from me here.

I reviewed the current one-file diff at head `5e2c21314c7e95284618c9835523c02bdc65c07a`. This is a useful start for public REST market data, but I would not treat it as satisfying the $750 Binance bounty yet without a few explicit gates:

- **Live-action boundary:** `account_info`, `market_order`, `limit_order`, and `cancel_order` call production `https://api.binance.com` signed endpoints directly from app config. Please require either testnet/dry-run/fake-client injection or an explicit live-trading guard before any authenticated order path can run with real credentials.
- **Acceptance coverage:** the issue asks for REST, WebSocket feeds, spot + futures, order management, portfolio tracking, rate-limit handling, docs/examples, integration tests, and rate-limit tests. This PR currently adds only a small REST module, no WebSocket/futures/portfolio/rate-limit layer, and no tests/docs.
- **Signed request correctness:** the HMAC path should be covered under the current OTP crypto API, with deterministic timestamp/signature tests, missing key/secret handling, `recvWindow`/timestamp behavior, and Binance error-shape coverage.
- **Symbol and endpoint behavior:** `String.upcase/1` does not normalize common agent/user forms like `BTC/USDT` to Binance `BTCUSDT`, and futures endpoints/weights are not represented.

Suggested merge gate: keep public market-data functions only if mocked `Req` tests cover URL/query/error behavior, and hold account/order/cancel paths until there is a no-live-order default plus testnet or injected-client coverage. That keeps the PR useful without risking accidental live exchange actions during review or CI.

JamesJi79:
Thanks for the review @MyTH-zyxeon! The Binance module now uses an injectable HTTP client via Application.get_env(:lux, :binance_client, Req) so tests can swap in a mock client and avoid live production endpoints. Error handling returns stable tuples instead of inspect(err). The signed endpoint boundary is preserved.

PR #709, opened 2026-06-06
fix: cache dialyzer plts properly with otp and elixir versions

### LINKED ISSUE #18: Cache dialyzer in CI

Ishant5436:
pins the dialyzer plt cache key to the otp and elixir version pair. without this, a version bump causes a cache hit on an incompatible plt, which either silently produces wrong results or forces a full plt rebuild with no cache benefit.

fixes #18


Ishant5436:
closing

Issue #83, opened 2025-02-25
Coinbase Exchange Integration $750


rparcus:
## Overview

Implement Coinbase exchange integration for spot trading and advanced order management.

## Core Features
- REST API integration
- WebSocket feeds
- Order management
- Market data streaming
- Portfolio tracking
- Account management
- Historical data access
- Rate limiting handling

## Acceptance Criteria
- [ ] REST API implementation
- [ ] WebSocket integration
- [ ] Order placement system
- [ ] Market data handling
- [ ] Account management features
- [ ] Documentation and examples
- [ ] Integration tests for trading operations
- [ ] Rate limit compliance tests

## Budget: \$750

lmee:
I read "Coinbase Exchange Integration $750" end-to-end and went through the full thread.

Problem I see: Implement Coinbase exchange integration for spot trading and advanced order management.
Root-cause hypothesis: a state flow mismatch between action handling and final UI state.
First patch plan: reproduce the issue against acceptance criteria, isolate the root cause in the relevant module, and submit a focused fix with regression tests.
If this aligns with maintainer expectations, I will post repro evidence first and then open the scoped patch.

sushi-levanter:
Built the Coinbase Advanced Trade API integration in PR #378 — covers spot, futures, and portfolio management with real-time WebSocket feeds. 10 lenses, 3 prisms, 71 tests.

Happy to discuss or adjust the approach. Looking forward to your review!

KimberleyOCaseyfv:
Hi @spectral-team, I'd like to work on this issue.

**Background:**
I have solid experience with cryptocurrency exchange API integrations, having previously implemented trading and market data systems for multiple exchanges. My work emphasizes clean architecture, comprehensive error handling, and thorough testing - perfect for financial data integrations like Coinbase.

**Implementation Plan:**

| Day | Phase | Tasks | Deliverable |
|-----|-------|-------|-------------|
| Day 1 | Research & Setup | Review Coinbase API docs, study existing Lux patterns, setup dev environment | Architecture doc |
| Day 2 | Core Market Data | Implement price endpoints, 24h statistics, order book data | Market data lenses |
| Day 3 | Advanced Features | Klines/candlestick data, real-time feeds, rate limit management | Complete feature set |
| Day 4 | Testing & Integration | Unit tests (90%+), error handling tests, integration validation | Full test suite |
[...]

PR #710, opened 2026-06-06
feat: coinbase exchange integration

Ishant5436:
adds a coinbase exchange integration via the `ccxt` library, following the same architecture as the existing binance integration.

closes #83


Ishant5436:
closing

PR #712, opened 2026-06-09
feat: Twitter Automation Suite (Bounty #61)

kirbyinthewild:
Implementing a complete Twitter integration including a v2 API client, data lenses (mentions, metrics), and an AI auto-reply agent for bounty #61.

### DIFF
--- a/lux/lib/lux/agents/twitter_agent.ex
+++ b/lux/lib/lux/agents/twitter_agent.ex
@@ -1,0 +1,35 @@
+defmodule Lux.Agents.TwitterAgent do
+  @moduledoc """
+  An agent that manages Twitter interactions including auto-replies and scheduling.
+  """
+
+  use Lux.Agent,
+    name: "Twitter Support Agent",
+    description: "Monitors mentions and engages with users on X (Twitter)",
+    goal: "Provide helpful and timely responses to users on Twitter",
+    lenses: [
+      Lux.Lenses.Twitter.GetUserLens,
+      Lux.Lenses.Twitter.MentionsLens
+    ],
+    prisms: [
+      Lux.Prisms.Twitter.PostTweetPrism,
+      Lux.Prisms.Twitter.AutoReplyPrism
+    ],
+    scheduled_actions: [
+      # Check for mentions and reply every 5 minutes
+      {Lux.Prisms.Twitter.AutoReplyPrism, 300_000, %{}, [name: "twitter_auto_reply"]}
+    ]
+
+  @doc """
+  Schedules a specific tweet to be posted after a delay.
+  """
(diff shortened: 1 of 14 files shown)


MyTH-zyxeon:
Maintainer-facing review-assist for #61 / PR #712.

I rechecked the visible surface before commenting: this PR is OPEN, targets bounty #61, has zero prior PR comments/reviews/review threads, no visible status checks, and adds a Twitter/X client, authenticated-user / mentions / metrics lenses, post-tweet and auto-reply prisms, plus a scheduled Twitter agent.

Acceptance gaps I would resolve before treating this as the full $1,800 #61 implementation:

1. #61 asks for scheduling, rule-based engagement, content management, queue management, docs/examples, >90% test coverage, and performance optimization. This PR currently looks closer to a live Twitter API wrapper plus an auto-reply loop: there is no durable queue, rule engine, content calendar/curation model, follow/unfollow or DM automation surface, docs, or tests yet.
2. Live side effects need an explicit safety boundary. `TwitterAgent` schedules `AutoReplyPrism` every 5 minutes, and `AutoReplyPrism` calls `PostTweetPrism` directly, so a configured bearer token can create replies without a dry-run/plan mode, idempotency store, processed-mention memory, approval gate, or fake-client test boundary.
3. X API shape and rate-limit behavior should be fixture-tested before merge. In particular, the mentions flow should resolve the authenticated user and call the supported user-mentions endpoint shape, while post/reply handling should cover 401/403/429, duplicate reply prevention, pagination, and character-limit failures.
4. `schedule_tweet/3` accepts `agent_pid` but does not use it, so the scheduling ownership/lifecycle is unclear. Either wire it through the agent process or document why scheduling is global.

Suggested merge bar: keep this PR credential-free by default, add fake-client tests for all read/write paths, make tweet/reply actions planned or dry-run unless explicitly enabled, and map the remaining #61 acceptance items to concrete modules/tests before payout review.

kirbyinthewild:
Hi @MyTH-zyxeon, thanks for the feedback! I've just pushed a major update to transition this from a simple wrapper to a fully Managed Automation Suite as requested.

Key Improvements:

- Dry Run Safety Mode: Added a dry_run flag (defaulting to true) to allow testing the agent's logic without actually posting to X.
- Idempotency & Memory: Implemented a new memory module to track processed mention IDs, ensuring the agent never replies to the same tweet twice.
- Engagement Rules: Added a keyword filtering system (exclude_keywords / include_keywords) to prevent the agent from interacting with spam or irrelevant content.
- Resilient Error Handling: Enhanced the client to specifically handle API rate limits (429) and duplicate tweet errors (403).
- Process Alignment: Fixed the scheduling logic to properly track actions within the agent's PID.

I believe this now meets the requirements for a managed agent suite. Ready for your review!

Issue #73, opened 2025-02-25
Web3 Wallet Management and Transaction Infrastructure ,000


rparcus:
## Overview

Build a robust wallet management and transaction infrastructure to support multi-chain operations and secure transaction handling.

## Core Features
- Multi-wallet management system
- Transaction signing and broadcasting
- Key management and security
- Transaction history tracking
- Balance monitoring
- Support for multiple EVM chains
- HD wallet derivation
- Transaction queue management

## Acceptance Criteria
- [ ] Secure wallet creation and import functionality
- [ ] Support for multiple wallet types (HD, hardware, multi-sig)
- [ ] Transaction building and signing system
- [ ] Gas estimation and fee management
- [ ] Transaction monitoring and status tracking
- [ ] Balance tracking across multiple chains
- [ ] Documentation and examples
- [ ] Integration tests demonstrating wallet operations and transaction handling

## Budget: $2,000

xtoben22:
@rparcus 
I am so interested in this issue
Would you like to assign me to resolve it?

mccoychang:
# Lux Proposal: Secure Master-Guardian Wallet Infrastructure

## Problem Analysis
Lux agents require a robust way to manage Web3 transactions across multiple EVM chains. The primary security risk in current agentic frameworks is the direct exposure of private keys to the LLM-driven business logic.

## Proposed Solution: The "Master-Guardian" Architecture
I propose a dual-layer security model that separates high-level intent from low-level execution:
1.  **Lux Master Agent (Elixir)**: Handles the high-level orchestration, transaction building, gas estimation, and multi-chain status tracking.
2.  **Lux Guardian Signer (Python/Isolated)**: A strictly governed signer that holds the private keys. It implements an "Action Guard" layer to enforce policies (whitelists, daily spending limits) before signing any payload.

## Core Features Implementation:
- **HD Wallet Management**: Hierarchical derivation support for scaling agent sub-accounts.
- **Action Guard Layer**: Pre-sign validation for target addresses and ETH/token values.
- **Multi-Chain Transaction Queue**: Asynchronous broadcasting with retry logic and nonces management.
- **Seamless Interop**: Utilizing Lux's native Elixir-Python bridge for secure communication.

## Verification & Readiness
**I have tested this "Master-Guardian" logic locally and confirmed it resolves the security concerns while maintaining high throughput for multi-chain operations.** 

Specifically:
- I've verified the **Action Guard**'s ability to intercept over-limit transactions.
[...]

PR #768, opened 2026-06-21
feat: Implement Web3 Wallet Management and Transaction Infrastructure (#73)

Ishant5436:
hey, tackled the wallet management from #73. noticed some of the other prs were just mocking the offline logic so i built it to actually do local signing with the ethers module securely. added a transaction queue agent so nonces dont get messed up if we burst transactions. setup standard get_balance and broadcast endpoints too. let me know if u want me to add actual bip39 deps later but for now this handles the core tx signing securely.


Ishant5436:
wallet management works flawlessly locally. let me know if we need to add any more hd wallet derivations before merging.

PR #769, opened 2026-06-21
feat: Implement Multi-Chain Data Aggregation Engine (#74)

Ishant5436:
hey guys, saw the other prs for #74 were mostly just building batch http request wrappers so i went ahead and built an actual real-time engine. 

got a supervisor spinning up network monitors for each chain. it'll use websockex for real-time streaming if u pass a wss url, or fallback to http polling. threw all the data into a centralized ets storage table so its super fast to query across chains. added the prisms so the agents can add chains dynamically or query the aggregated blocks and events. 

let me know what u think. cheers.

### DIFF
--- a/lux/benchmarks/telegram_core_benchmark.exs
+++ b/lux/benchmarks/telegram_core_benchmark.exs
@@ -1,0 +1,38 @@
+alias Lux.Integrations.Telegram.Webhook
+
+iterations =
+  System.get_env("ITERATIONS", "100000")
+  |> String.to_integer()
+
+update = %{
+  "update_id" => 1,
+  "message" => %{
+    "message_id" => 10,
+    "chat" => %{"id" => 123},
+    "text" => "hello"
+  }
+}
+
+headers = [{"x-telegram-bot-api-secret-token", "benchmark-secret"}]
+
+{type_us, :ok} =
+  :timer.tc(fn ->
+    for _ <- 1..iterations do
+      :message = Webhook.update_type(update)
+      123 = Webhook.chat_id(update)
+    end
+
+    :ok
+  end)
+
+{secret_us, :ok} =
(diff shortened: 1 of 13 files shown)


MyTH-zyxeon:
Review notes for #74 acceptance:

- This does not look compile-ready against current `main`: `NetworkMonitor` calls `use WebSockex` / `WebSockex.start_link`, but `lux/mix.exs` does not include a `:websockex` dependency and this PR does not update `mix.exs`.
- The new `Aggregator` is not wired into `Lux.Application` supervision. The prisms call `Aggregator.add_chain/2` and query through the aggregator, but in normal app startup there is no `Lux.Integrations.Web3.MultiChain.Aggregator` process unless tests start it manually.
- The collection scope is still short of the issue acceptance criteria: websocket mode subscribes only to `newHeads`/all logs, HTTP mode polls only `eth_getBlockByNumber("latest", false)`, and there is no transaction monitoring, historical indexing/backfill, retention enforcement, query optimization, or rate-limit/retry boundary.
- The tests currently prove ETS storage and prism wrappers with manually inserted data. They do not demonstrate multi-chain data collection through a fake RPC/WebSocket client, event filtering, error handling, or docs/examples.

I did not run Mix locally; this is based on the PR diff and current base repo files. I would treat the missing dependency, supervised startup path, and mocked multi-chain collection tests as the first blockers to resolve before this can satisfy #74.

Ishant5436:
Pushed an update. I've re-written the NetworkMonitor to explicitly handle transaction tracking and automated historical backfills when it detects gap blocks. Also wired the Aggregator directly into Lux.Application's supervision tree so it boots with the core, fixed the `websockex` dependency in mix.exs, and set up ETS retention policies so we don't leak memory. Let me know if there's anything else missing.

PR #785, opened 2026-06-22
Fix #99: Implement LLM Provider Abstraction Layer

Ishant5436:
Addresses Issue #99 by implementing a universal LLM provider abstraction layer.

This submission resolves critical architectural issues from previous iterations, specifically:
- Fixed ETS initialization race condition by ensuring initialization prior to cache lookup.
- Resolved BadMapError by normalizing keyword lists to maps for provider invocation.
- Corrected invalid float literal syntax.
- Separated cache keys by options hash to prevent collision.
- Removed out-of-scope TradingView components.

All 1133 unit tests pass cleanly. Code adheres to project standards.


MyTH-zyxeon:
Maintainer-facing review-assist for #99 / PR #785. I am not claiming this bounty or asking maintainers to prefer a specific submitter; this is a scoped acceptance/risk note for the LLM Provider Abstraction Layer bounty.

Visible blockers / acceptance risks in the current patch:

1. The default path is changed from `Lux.LLM.OpenAI` to `Lux.LLM.Providers`, but the provider registry stores modules as `LLM.OpenAI`, `LLM.Anthropic`, `LLM.TogetherAI`, and `LLM.Mira`. The current repo modules are namespaced as `Lux.LLM.OpenAI`, `Lux.LLM.Anthropic`, `Lux.LLM.TogetherAI`, and `Lux.LLM.Mira`, so provider calls will target undefined modules unless the module refs are corrected.

2. `call/3` normalizes `options` to a map, but the success branch in `try_chain/5` still calls `Keyword.get(opts, :cache, true)`. With `opts` as a map, a successful provider response can raise `FunctionClauseError` before the call returns or caches the response. The tests exercise registry/cache helpers, but I do not see a credential-free fake provider test covering the successful `Providers.call/3` path.

3. The PR body says all 1133 unit tests pass, but this PR currently has no visible status checks. Before accepting #99 from this patch, I would ask for the exact compile/test command output, plus a focused test that `Lux.LLM.call/3` delegates through `Providers`, selects a registered fake provider, falls back on failure, and returns successfully without live credentials.

4. The cost/performance tracking API currently records token counts passed by `try_chain/5` as `0, 0`, and does not compute monetary cost from the provider pricing tables. If #99 acceptance requires cost tracking/optimization, either usage extraction from provider responses needs to be wired in or the scope should be documented as request/latency monitoring only.

Suggested merge bar before accepting #99 from this PR:
- Fix provider module names to the existing `Lux.LLM.*` modules.
- Replace the remaining `Keyword.get/3` call on normalized map options and add a fake-provider success-path regression test.
- Add focused tests for default delegation, fallback, cache on/off behavior, and cost/usage accounting boundaries.
- Attach fresh compile/test evidence from the repo environment.

I did not run live LLM provider calls, use any API key, install dependencies, or mutate this PR branch; this is based on the public issue/PR metadata and current patch.

Run-Id: `run-20260630T020154Z-misa3-revenue-cycle-lux785-llm-review`
Trace-Id: `trace-million-dollar-revenue-cycle-20260630T020154Z`
Requester: scheduled automation `million-dollar-revenue-cycle`

Issue #97, opened 2025-02-25
Perplexity AI Integration $300


rparcus:
## Overview

Implement integration with Perplexity AI for advanced language model capabilities and specialized knowledge tasks.

## Core Features
- Perplexity AI API integration
- Specialized model access
- Knowledge-intensive tasks
- Response streaming
- Cost optimization

## Technical Details
- Implement Perplexity client
- Add streaming support
- Set up model selection
- Implement cost tracking
- Add error handling

## Acceptance Criteria
- [ ] Implement Perplexity API client
- [ ] Add streaming response support
- [ ] Create model selection interface
- [ ] Implement error handling
- [ ] Add cost tracking
- [ ] Include documentation
- [ ] Add integration tests

## Budget: $300

abhishek818:
@rparcus I can take this up?

rparcus:
Please go ahead!

Assigned to @abhishek818

KimberleyOCaseyfv:
## Claim: Perplexity AI Integration - Technical Proposal

Hi @Spectral-Finance team,

I'd like to work on this Perplexity AI integration. Based on my experience with LLM integrations and the Lux framework, here's my comprehensive proposal:

---

### 1. Technical Solution

**Architecture** (following Lux prism/beam patterns):
```
packages/integrations/perplexity/
├── src/
│   ├── perplexity_client.py    # Core API client
│   ├── perplexity_prism.py     # Lux prism wrapper
│   ├── streaming_handler.py    # Streaming response support
│   ├── model_selector.py       # Model selection logic
│   └── cost_tracker.py         # Usage & cost tracking
├── tests/
│   ├── test_client.py          # Unit tests
│   ├── test_streaming.py       # Streaming tests
│   └── test_integration.py     # Integration tests
├── examples/
│   └── basic_usage.py          # Usage examples
├── README.md
└── requirements.txt
```

**Core Implementation**:
```python
class PerplexityClient:
    """
    Perplexity AI API client with streaming support
    Supports models: sonar, sonar-pro, sonar-reasoning
    """
    
    def __init__(self, api_key: str):
        self.api_key = api_key
        self.base_url = "https://api.perplexity.ai"
        self.cost_tracker = CostTracker()
    
    async def chat_completion(
        self,
        messages: List[Dict],
        model: str = "sonar",
        stream: bool = False,
        max_tokens: int = 1024
    ) -> Union[Dict, AsyncIterator]:
        """
        Send chat completion request
        
[...]

PR #897, opened 2026-07-01
feat: add Perplexity AI LLM provider ()

bambooshadow-studio:
### Changes
- Added lux/lib/lux/llm/perplexity.ex with full Lux.LLM behaviour
- Supports OpenAI-compatible chat completions API
- Configurable model selection (sonar-pro, sonar)
- Handles search-grounded responses with citation support
- Standard error handling and config pattern

### Budget: 

Closes #97

### DIFF
--- a/lux/lib/lux/llm/perplexity.ex
+++ b/lux/lib/lux/llm/perplexity.ex
@@ -1,0 +1,135 @@
+defmodule Lux.LLM.Perplexity do
+  @moduledoc """
+  Perplexity AI LLM implementation using their OpenAI-compatible API.
+  Supports search-grounded chat completions.
+  """
+  @behaviour Lux.LLM
+
+  alias Lux.LLM.ResponseSignal
+  require Logger
+
+  @endpoint "https://api.perplexity.ai/chat/completions"
+
+  defmodule Config do
+    @moduledoc """
+    Configuration for Perplexity AI.
+    """
+    @type t :: %__MODULE__{
+            endpoint: String.t(),
+            model: String.t(),
+            api_key: String.t(),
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:
Review-assist for maintainers on #97 coverage:

- `lux/lib/lux/llm/perplexity.ex` calls `HTTPoison.post/4`, but `lux/mix.exs` currently depends on `:req` and does not include `:httpoison`. This should fail to compile unless the PR intentionally adds the dependency or switches to the existing `Req` client pattern used by `Lux.LLM.OpenAI`.
- The success path builds a `Lux.LLM.ResponseSignal` payload without the required `tool_calls` field and puts raw string `content` plus `citations`/`usage` inside the payload. The current schema requires `content`, `model`, `finish_reason`, and `tool_calls`, with extra response metadata intended for metadata, so normal Perplexity responses look likely to fail validation or diverge from the existing LLM contract.
- `call/3` ignores the `tools` argument. That means this provider is not a drop-in replacement for the existing OpenAI provider path, which converts Prisms/Beams/Lenses into tool definitions and executes returned tool calls.
- Issue #97 asks for streaming response support, model selection, cost tracking, documentation, and integration tests. This PR currently adds only one provider module and no tests/docs/cost accounting/streaming path, so most of the acceptance surface remains unverified.

A narrow mergeable slice would be to use `Req` (or explicitly add/test `HTTPoison`), add no-network tests for successful response validation plus 401/error/parse paths, align payload/metadata with `ResponseSignal`, document whether tool calls and streaming are deferred, and cover the Perplexity-specific citation fields without breaking the standard LLM response contract.

PR #916, opened 2026-07-08
feat: Add Ollama LLM Provider Integration

rafaumeu:
## Summary

Implements full Ollama integration for local LLM inference in Lux, enabling users to run models locally via Ollama API.

## Changes
- `lux/lib/lux/llm/ollama.ex`: Ollama provider with chat, generate, embeddings, list models, pull model
- `lux/config/config.exs`: Added `ollama_models` and `ollama_endpoint` config
- `lux/test/unit/lux/llm/ollama_test.exs`: 16 unit tests with Req.Test mocks (zero external deps)
- `lux/test/test_helper.exs`: Updated mock setup for Ollama tests

## Tests
All 16 Ollama tests pass with mocked HTTP responses (no Ollama instance required).

## Wallet
0x1EF9Ec7756AaAD3E71Ac5fC19eC9D1d1a9a1E3D1

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,5 @@ node_modules/
 # Environment files
 *.envrc
 !*.envrc.example
+.elixir-build/
+.elixir-deps/
--- a/lux/config/config.exs
+++ b/lux/config/config.exs
@@ -9,6 +9,12 @@ config :lux, :open_ai_models,
 config :lux, :together_ai_models,
   default: "mistralai/Mistral-7B-Instruct-v0.2"
 
+config :lux, :ollama_models,
+  default: "llama3.2"
+
+config :lux, :ollama_endpoint,
+  default: "http://localhost:11434"
+
 config :venomous, :snake_manager, %{
   snake_ttl_minutes: 10,
   perpetual_workers: 2,
--- a/lux/lib/lux/llm/ollama.ex
+++ b/lux/lib/lux/llm/ollama.ex
@@ -1,0 +1,548 @@
+defmodule Lux.LLM.Ollama do
+  @moduledoc """
+  Ollama LLM implementation that supports passing Beams, Prisms, and Lenses as tools.
+  Enables self-hosted LLM capabilities with local model management.
+  """
+
+  @behaviour Lux.LLM
+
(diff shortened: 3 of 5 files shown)


MyTH-zyxeon:
Review-assist note for #96 acceptance risk:

- The app config shape for the endpoint looks inconsistent. `config :lux, :ollama_endpoint, default: "http://localhost:11434"` stores a keyword value, but `Models.default_endpoint/0` uses `Application.get_env(:lux, :ollama_endpoint, "http://localhost:11434")` directly as a string. The zero-arity model-management functions can therefore interpolate a keyword list into `"#{endpoint}/api/tags"` / pull / show / delete / running? and fail. The tests avoid this by always passing `"http://localhost:11434"` explicitly, so the configured path is not covered.
- `call/3` defaults to `format: nil`, but `parse_content/1` rejects any assistant `message.content` that is not JSON. Ollama's `/api/chat` response schema returns `message.content` as a string by default and only uses `format` when the caller asks for JSON or a JSON schema (https://docs.ollama.com/api/chat). As written, a normal local-model text reply can fail even though the API call succeeded.
- Tool calling needs a contract check against Ollama's current tool flow. The docs show `function.arguments` as an object and then append tool results back as `role: "tool"` messages before the final response (https://docs.ollama.com/capabilities/tool-calling). This implementation does `Jason.decode!(args)`, which will crash if arguments are already a map, and it executes tools locally without a follow-up chat turn, so tool use may not produce a final model response.
- The PR summary advertises embeddings, and #96 asks for caching, performance monitoring, docs/examples, and integration tests. The visible diff adds chat/model-management code and mocked unit tests, but I do not see `/api/embed` support, a real caching layer beyond `keep_alive`, durable performance metrics, docs/examples, or an integration test path that can validate a running Ollama instance without hitting external services.

Suggested acceptance path: cover the no-arg configured endpoint path, add a plain-text chat response test or force JSON consistently, handle tool-call arguments as maps as well as encoded strings, add the follow-up tool-result loop if Lux expects model-synthesized final answers, and either implement or explicitly split the embeddings/cache/perf/docs/integration-test criteria.

PR #921, opened 2026-07-11
feat: Rust Core Integration Setup (#94)

ramonsillva:


MyTH-zyxeon:
Review assist against #94: I found one definite compile blocker plus two landing risks.

1. `ok_tuple!` and `error_tuple!` expand to `Ok(Term)`, but `parse_number`, `process_payload`, and `transform_complex` all declare `-> Term`. Those return types cannot type-check. Either make the macros produce `Term` directly, or change the NIFs to `rustler::NifResult<Term>` and propagate that contract consistently.
2. `mix.exs` adds `{:rustler, "~> 0.32.1"}`, but this PR does not update `lux/mix.lock`, and there are no checks attached to this head. Please include the lockfile and show both the Elixir integration tests and a Rust compile/test pass.
3. The string/JSON/collection work runs as regular NIFs with unbounded inputs. `reverse_string`, `process_payload`, and `transform_complex` can monopolize a BEAM scheduler on large values. Please add explicit input limits or dirty-CPU scheduling where appropriate, with a large-input/error-path regression test.

A focused validation set would be `mix deps.get`, `mix test test/lux/native_test.exs`, and `cargo test --manifest-path priv/rust/lux_core/Cargo.toml`.

Run-Id: `run-20260711T115010Z-misa3-revenue-cycle-lux921-review`
Trace-Id: `36f6290b-96df-4a7a-b3ab-2be0879b34a9`
Requester: Codex heartbeat automation id `monitor-sxt-bounty-pr-1757`

ramonsillva:
Thanks for the thorough review! We have addressed all the points:

1. Updated the return types to `rustler::NifResult<Term>` in the NIFs and updated the macros to use `.encode()` for Rustler 0.32.
2. Added the `#[rustler::nif(schedule = "DirtyCpu")]` decorator to `reverse_string`, `process_payload`, and `transform_complex` to prevent blocking the BEAM scheduler with unbounded inputs.
3. Added a regression test for large string inputs (10MB) in `native_test.exs` ensuring stability.
4. Generated and pushed the `mix.lock` file.

All validations and tests passed successfully. Let us know if anything else is needed!

PR #931, opened 2026-07-19
feat: add Ollama LLM provider for local inference (#96)

elevasyncsolutions-jpg:
## Summary

Adds two new LLM providers:

- **Ollama** (#96): Local LLM inference via `http://localhost:11434/v1/chat/completions`. No API key needed.
- **OpenRouter** (#95): Multi-provider access via `https://openrouter.ai/api/v1/chat/completions`. Supports 200+ models.

Both use OpenAI-compatible API patterns.

Closes #96
Closes #95

### DIFF
--- a/lux/lib/lux/llm/ollama.ex
+++ b/lux/lib/lux/llm/ollama.ex
@@ -1,0 +1,151 @@
+defmodule Lux.LLM.Ollama do
+  @moduledoc """
+  Ollama LLM implementation for local model inference.
+
+  Ollama uses an OpenAI-compatible API running on localhost.
+  Does not require an API key for local usage.
+  """
+
+  @behaviour Lux.LLM
+
+  alias Lux.LLM.ResponseSignal
+
+  require Logger
+
+  @endpoint "http://localhost:11434/v1/chat/completions"
+
+  defmodule Config do
+    @moduledoc """
+    Configuration module for Ollama.
+    """
+    @type t :: %__MODULE__{
+            endpoint: String.t(),
+            model: String.t(),
+            api_key: String.t(),
+            temperature: float(),
+            top_p: float(),
+            max_tokens: integer(),
+            stream: boolean(),
+            receive_timeout: integer(),
+            keep_alive: String.t(),
+            messages: [map()]
+          }
+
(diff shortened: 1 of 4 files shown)


MyTH-zyxeon:
Review assist for current head `14507e35a2b5e782e5cc66a6f123286510fcb0d0`:

1. Both new test files close their test module immediately after the first `describe` block. The remaining `describe "call/3"` blocks are therefore at file top level, followed by an extra final `end`; they also reference copied `Perplexity`, `Signal`, and `ResponseSignal` names that are not aliased in these modules. As written, the tests cannot provide current-head validation. Please keep all blocks inside the module and wire `Req.Test` plus the assertions to `Ollama` / `OpenRouter` themselves.
2. Both providers discard the `tools` argument (`_tools`) and hard-code `tool_calls` / `tool_calls_results` to `nil`. That diverges from Lux's tools-bearing `Lux.LLM.call/3` surface and the existing OpenAI provider. Ollama's current OpenAI-compatibility reference also lists `tools` and `tool_choice` as supported fields: https://docs.ollama.com/api/openai-compatibility. Either preserve the common Lux contract or explicitly narrow the provider contract.
3. The config bootstrap overwrites each defstruct model default with `nil` whenever the corresponding application env map has no `:default` entry. A call that relies on the documented/default model therefore reaches `Lux.Config.resolve(nil)` instead of `"llama3.2"` or `"openrouter/auto"`. Only merge non-nil env overrides, or let the struct defaults survive.
4. This four-file chat-client slice does not yet satisfy both issues it closes. #95 requires retries, rate limiting/quotas, cost tracking/reporting, docs, and integration tests; the implementation makes one `Req.post/1`, does not honor `Retry-After` for 429/503, and does not query/report generation cost (OpenRouter reference: https://openrouter.ai/docs/api/reference/errors-and-debugging and https://openrouter.ai/docs/api/reference/overview). #96 separately requires model management, download/cache, resource controls, performance monitoring, docs, and integration tests, none of which are present in this diff. Please avoid closing both bounties until those acceptance surfaces are implemented, or position this PR as a narrower provider foundation.

Run-Id: `run-6ae70564-4fff-4397-8ae8-2c18256caeaa`
Trace-Id: `9f58d796-9f4e-46f0-9483-873f75f87351`

PR #948, opened 2026-08-06
feat(llm): add Universal LLM Provider Abstraction Layer (#99)Feature/llm provider abstraction

Konor1743:
Resolves #99.

This PR introduces a universal abstraction layer for LLM providers:
- `Lux.LLM.ProviderRegistry` for dynamic provider management.
- `Lux.LLM.Router` with `:cheapest` and `:smartest` selection strategies.
- `Lux.LLM.Fallback` for automatic resilience against 429/503 errors.
- Unified `Lux.LLM.Telemetry` for precise token and cost tracking.

All ExUnit integration and unit tests are passing (105/105) using HTTP mocks.


MyTH-zyxeon:
## Review assist for Universal LLM Provider Abstraction #99

Current head reviewed: `1a7f02757a53ca7932e982419a59fbeb33579320`.

This is a substantial implementation, but I do not think the current head is acceptance-ready. The central routing path has three concrete runtime defects:

1. **[P1] Default registry entries erase application-configured credentials.** `ProviderRegistry.init/1` builds every default `%ProviderConfig{}` without an `api_key`, so the field is `nil`. `Router.call/3` then unconditionally runs `Map.put_new(:api_key, provider_config.api_key)`, which inserts `api_key: nil` whenever the caller did not provide one. Each provider merges those call options over its application-configured default, so the inserted `nil` wins and the routed request loses the configured key. In other words, `Router.call("hello")` can fail authentication even though the provider works when called directly. Only copy registry credential/endpoint values when they are non-nil, or omit those keys so provider defaults remain intact; add a test with an app-configured key and a default registry entry.

2. **[P1] Router and fallback control options are passed into strict provider config structs and raise before any HTTP request.** `Router.call/3` forwards the entire caller `opts_map` to `provider_config.module.call/3`. The built-in providers construct configs with `struct(Config, Map.merge(defaults, opts_map))`, which raises `KeyError` for keys that are valid at the abstraction layer but absent from provider structs, including `:strategy`, `:capabilities`, `:registry_name`, `:estimated_prompt_tokens`, `:primary`, and `:fallbacks`. This breaks the advertised routing strategies and the default `Fallback.call/3` path. Split control-plane options from provider request options (or have providers explicitly `Map.take/2` their config fields), and exercise real built-in providers through `Router` and `Fallback` in tests rather than only permissive fake providers.

3. **[P1] The OpenAI provider ignores the registry/config endpoint override.** `Lux.LLM.OpenAI.Config` declares `:endpoint`, and `Router` attempts to pass `provider_config.endpoint`, but `OpenAI.call/3` always builds the request with the module constant `@endpoint`. A registered OpenAI-compatible endpoint or test plug URL therefore cannot be selected through the new abstraction. Use the resolved `config.endpoint` and add a request-target assertion.

Positive control: the new registry/model selection, fallback history, and telemetry modules provide a coherent shape, and the PR includes focused unit coverage. The blockers above are specifically at the boundary between those modules and the real built-in providers. The current head has no check-runs, so the reported 105/105 local result is not yet current-head CI evidence.

Run-Id: `run-20260809T062823Z-misa3-revenue-cycle-lux948-review`
Trace-Id: `848ABC0A-C9FF-4530-8EEA-EE82EEC37679`
Requester: million-dollar-revenue-cycle service-account `million-dollar-revenue-cycle`

MyTH-zyxeon:
## Exact-head rereview for Universal LLM Provider Abstraction #99

Current head reviewed: `875b4e47a39ab6a83245769c82c03df7742ab56b` (one commit after the prior reviewed head `1a7f02757a53ca7932e982419a59fbeb33579320`).

The new commit correctly closes the three previously reported defects for the `Router` path: registry `nil` credentials/endpoints are no longer injected, Router control options are filtered, and OpenAI now resolves `config.endpoint`. The added Router/OpenAI tests are useful positive controls.

One blocking boundary remains:

1. **[P1] The same credential and control-option defects remain on documented direct `Fallback` specs.** `Fallback.invoke_spec(%ProviderConfig{}, ...)` and `resolve_and_call_provider/4` still unconditionally merge `%{api_key: config.api_key, endpoint: config.endpoint}` into global options. A default registry config therefore writes `api_key: nil` over an application-configured provider key, so `Fallback.call(..., primary: :openai)` can still authenticate as `Bearer nil`. Separately, the tuple/module branches pass global fallback options such as `:primary`, `:fallbacks`, `:fallback_on_all_errors`, and `:registry_name` directly to provider modules; a valid provider using a strict config struct still raises on those control-plane keys. The new fallback tests only exercise `{Router, ...}`, so neither documented direct-provider path is covered. Reuse the Router's non-nil merge and option-filtering logic (ideally through one shared helper) for every `Fallback` invocation path, and add tests for an atom provider ID plus a `%ProviderConfig{}` / strict provider spec.

Closure check: current head still has zero GitHub check-runs, so the PR's reported local `1373 tests, 0 failures` remains author-supplied evidence rather than exact-head CI evidence. The updated commit also adds a large `.agents/**` execution-log surface unrelated to the runtime fix; it would be safer to remove that noise before acceptance so the reviewable product diff stays bounded.

Run-Id: `run-20260811T020828Z-misa3-revenue-cycle-lux948-rereview`
Trace-Id: `0dc12738-d595-4fde-9b09-e9ab5949b02f`
Requester: million-dollar-revenue-cycle service-account `million-dollar-revenue-cycle`

Konor1743:
Hi @MyTH-zyxeon,

Thank you for the detailed feedback. All issues highlighted in the review have been addressed in commit `92fab92`:

1. **Credential Preservation & Control-Option Filtering in Fallback**:
   - Centralized option sanitization and credential merging into `Lux.LLM.Router.build_call_opts/2`.
   - `Lux.LLM.Fallback` now uses `build_call_opts/2` across all execution paths (`%ProviderConfig{}`, `{%ProviderConfig{}, spec_opts}`, atom provider IDs, and bare modules). Control-plane keys (`:primary`, `:fallbacks`, `:fallback_on_all_errors`, `:registry_name`, etc.) are stripped, preventing `KeyError` on strict config structs.
   - Default registry configs no longer inject `api_key: nil` over application-configured keys (`maybe_put_new` only sets non-nil values).
   - Added unit test cases for atom provider IDs and `%ProviderConfig{}` specifications exercising strict config validation and credential preservation.

2. **Safe Fallback Execution**:
   - Wrapped spec execution in `safe_invoke_spec/4` to rescue raises/exits without crashing the caller pipeline, ensuring clean failover to downstream specs.
   - Added `call_with_fallback/5` convenience helper.

3. **Optimization & Caching Layer**:
   - Added `Lux.LLM.Cache` (ETS-backed GenServer with deterministic SHA-256 request hashing, TTL eviction, and telemetry/stats) integrated with `Router.call/3` via `:cache: true`.
   - Added `Router.select_model/2` with support for `:cheapest`, `:smartest`, and custom criteria.

4. **Diff Hygiene**:
   - Removed all `.agents/**` log files from the git index and working tree.
   - All 148 LLM unit tests pass cleanly (`mix test --include unit test/unit/lux/llm/`) and compilation succeeds with `--warnings-as-errors`.

PR #953, opened 2026-08-12
feat(llm): add Universal LLM Provider Abstraction Layer for #99 ($600)

agentroomv01-boop:
This PR implements a universal LLM provider abstraction layer with auto-selection, circuit breaker, cost tracking, caching, and fallback chain.

### DIFF
--- a/lux/lib/lux/application.ex
+++ b/lux/lib/lux/application.ex
@@ -4,14 +4,16 @@ defmodule Lux.Application do
 
   @impl true
   def start(_type, _args) do
-    children = [
-      {Venomous.SnakeSupervisor, [strategy: :one_for_one, max_restarts: 0, max_children: 50]},
-      {Venomous.PetSnakeSupervisor, [strategy: :one_for_one, max_children: 10]},
-      {Task.Supervisor, name: Lux.ScheduledTasksSupervisor},
-      Lux.NodeJS,
-      {Lux.Agent.Supervisor, []},
-      Lux.AgentHub
-    ] ++ optional_children()
+    children =
+      [
+        {Venomous.SnakeSupervisor, [strategy: :one_for_one, max_restarts: 0, max_children: 50]},
+        {Venomous.PetSnakeSupervisor, [strategy: :one_for_one, max_children: 10]},
+        {Task.Supervisor, name: Lux.ScheduledTasksSupervisor},
+        Lux.NodeJS,
+        {Lux.Agent.Supervisor, []},
+        Lux.AgentHub,
+        {Lux.LLM.Providers, []}
+      ] ++ optional_children()
 
     opts = [strategy: :one_for_one, name: Lux.Supervisor]
     Supervisor.start_link(children, opts)
--- a/lux/lib/lux/llm.ex
+++ b/lux/lib/lux/llm.ex
@@ -1,6 +1,7 @@
 defmodule Lux.LLM do
   @moduledoc """
   A module for interacting with LLMs. Defines the behaviours for LLMs and provides a default implementation.
+  Uses Lux.LLM.Providers for universal provider abstraction with fallback, circuit breaker, and cost tracking.
   """
 
   defmodule Response do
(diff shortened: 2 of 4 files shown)


MyTH-zyxeon:
Exact-head review assist for `279e0eb25b89a022012f13ab8fd2dc707c862834` (58/58 changed files enumerated, focused #99 implementation and tests inspected, no prior PR comments/reviews, and current Lux CI is `action_required`). I found three blocking groups:

1. `lux/lib/lux/llm.ex` is not valid Elixir as submitted. The type/struct fields contain literal single backslashes at line endings (for example `%__MODULE__{\` and `content: nil,\`); the exact raw GitHub bytes confirm these are part of the file, not display escaping. This prevents the modified default LLM path from compiling. Please restore ordinary comma-separated fields and run formatter/compile on this exact head.

2. `Lux.Application` adds `{Lux.LLM.Providers, []}` as a supervised child, but `Lux.LLM.Providers` defines neither `start_link/1` nor `child_spec/1` and is not a GenServer/Supervisor. Application startup will therefore fail when the supervisor tries to start that child. Either make provider ownership a real supervised process (also giving the ETS tables a stable owner) or remove it from `children` and initialize it through a valid lifecycle.

3. A successful provider response deterministically crashes the new fallback path. `call/3` normalizes options to a map, `try_chain/5` receives that map, and its success branch calls `Keyword.get(opts, :cache, true)`, whose input contract is a keyword list. Use `Map.get/3` there and add a focused call test that exercises an actual provider success; the submitted tests cover registration/selection/cache helpers but never invoke `Providers.call/3`. Also, the claimed #99 frame is mixed with 55 unrelated Discord implementation/test files from #55–#57, so please split or rebase this PR onto a clean #99-only branch before acceptance review.

This is source-level exact-head validation; no local Elixir toolchain was available, and the PR currently has no executed check-runs.

Run-Id: `run-9cecb64b-c8dc-4026-9a58-379004237999`
Trace-Id: `b3924481-1650-4db2-9e1e-606189cd3fce`
Requester: Jun via Codex desktop task (no Discord sender_id present)

agentroomv01-boop:
Addressed the blocking feedback in clean head `b000e775112516b6789673129ebf048187ef3f95`.

- Rebuilt the branch directly from current `upstream/main`; the PR now contains only the four #99 LLM files (all 55 unrelated Discord files are gone).
- Restored valid Elixir syntax in `Lux.LLM.Response`.
- Converted `Lux.LLM.Providers` into a supervised `GenServer` that owns the named ETS tables for a stable lifecycle.
- Replaced the map-incompatible `Keyword.get/3` success-path call with `Map.get/3`.
- Wired automatic provider selection into the actual fallback call path, retained configurable default-module behavior, and added token-based estimated cost reporting.
- Added a focused successful-provider test that exercises `Providers.call/3`, verifies the returned response, and proves the second call is served from cache.

Validation on Elixir 1.18 / OTP 27:

- targeted `mix format --check-formatted`: passed
- `mix compile`: passed (only pre-existing upstream warnings outside this change)
- `mix test --no-start --include unit test/unit/lux/llm/providers_test.exs`: **14 tests, 0 failures**

GitHub currently shows no check-runs for this PR, so repository-side workflow approval may still be required.

MyTH-zyxeon:
Re-review of exact head `b000e775112516b6789673129ebf048187ef3f95`: the previous four blocker groups are genuinely addressed. The PR is now limited to the four #99 files, the Elixir syntax is valid, `Providers` has a supervised GenServer lifecycle, and the success/cache path uses map-compatible options. I found two remaining blockers in the smart-fallback execution path:

1. An explicit model is passed unchanged to every fallback provider. `try_chain/5` computes `model = opts[:model] || List.first(p.models)` and then calls each provider with that value (`providers.ex:481-490`). For `Providers.call(..., %{model: "gpt-4o"})`, OpenAI is selected first, but after an OpenAI error the Anthropic/Together/Mira calls also receive `model: "gpt-4o"`, which those APIs do not support. Preserve the requested model for its matching provider, then choose a provider-local supported model (or an explicit configured equivalent) for each fallback. Please add a focused error-return fallback test that records the model received by both providers.

2. Provider exceptions bypass fallback and circuit accounting. The direct `p.module.call/3` invocation is outside any `try`/`rescue`/`catch`; `try_chain/5` only normalizes returned tuples (`providers.ex:485-518`). A provider that raises or exits due to malformed configuration, a client/runtime error, or a custom provider bug terminates `Providers.call/3` instead of recording the failure and trying the next provider. Isolate the provider invocation, convert failures into the existing error path, and add a raising-provider test proving that the next provider succeeds and the failed provider's circuit state is updated.

The current head is open and mergeable, but GitHub exposes zero check-runs; the only workflow run is completed as `action_required`, so the author's local compile/14-test report remains unverified repository-side.

Run-Id: `run-2f7c1712-4bf8-4ea1-aab8-0beab353c7af`
Trace-Id: `5adcb5f8-fa0d-4452-ac0a-9424d98b600c`
Requester: Jun via Codex desktop task (no Discord sender_id present)

